PR #8, opened 2017-05-06
Pretty logs, address option, update notifier

delvedor:
This pr is to prepare the direct integration of the cli with Fastify, once we approve this I'll update Fastify by adding the following code to the `package.json`:
```json
  "bin": {
    "fastify": "./node_modules/.bin/fastify"
  },
```
In this way the users can use the cli out of the box.

What is inside this pr:
- Added pretty logs with [pino-colada](https://github.com/lrlna/pino-colada)
- Added [address](https://github.com/fastify/fastify/blob/master/docs/Server-Methods.md#listen) option
- Added [update notifier](https://github.com/yeoman/update-notifier) for fastify

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,23 +2,36 @@
 
 'use strict'
 
-const minimist = require('minimist')
-const Fastify = require('fastify')
 const path = require('path')
-const pino = require('pino')
-const pump = require('pump')
 const fs = require('fs')
+const assert = require('assert')
+
+const updateNotifier = require('update-notifier')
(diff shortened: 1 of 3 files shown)


delvedor:
Yeah, you are right.
If we install `fastify-cli` as dependency in Fastify and we add it in `bin` will work only if the user runs it via npm.

> npm layout is not-deterministic, so you'll need to do require('fastify-cli')() or something similar.

I didn't understand this.

PR #12, opened 2017-09-21
Add prefix. Refactor to run multiple tests

allevo:
Refactorize in order to return fastify instance. So the test can start and stop the instance and run multiple tests.

Add `prefix` command line parameter


mcollina:
The tests are failing on travis.

allevo:
@mcollina sorry my bad, old `fastify` version that not supports `prefix` option

PR #15, opened 2017-10-12
Socket listen

allevo:
Add socket option.
Add `pre-commit` package as devDependency

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -77,7 +77,7 @@ function runFastify (opts) {
 
   if (opts.address) {
     fastify.listen(opts.port, opts.address, listen)
-  } if (opts.socket) {
+  } else if (opts.socket) {
     fastify.listen(opts.socket, listen)
   } else {
     fastify.listen(opts.port, listen)
--- a/cli.js
+++ b/cli.js
@@ -85,7 +85,11 @@ function runFastify (opts) {
 
   function listen (err) {
     assert.ifError(err)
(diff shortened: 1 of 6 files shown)


test.js
```diff
@@ -72,6 +75,36 @@ test('should start the server at the given prefix', t => {
   })
 })
 
+test('should start fastify at given socket path', t => {
+  t.plan(2)
+
+  const sockFile = path.join(__dirname, '/test.sock')
```
delvedor: Probably `path.resolve` is better :)
Checkout how we used it [here](https://github.com/fastify/point-of-view/blob/master/index.js) :)

allevo: done

cli.js
```diff
@@ -77,6 +77,8 @@ function runFastify (opts) {
 
   if (opts.address) {
     fastify.listen(opts.port, opts.address, listen)
+  } if (opts.socket) {
```
delvedor: `else if`?

allevo:
Should we throw an error if the user specify `address`  or `port` with `socket`?

cli.js
allevo: Wow! strange... `standard` says nothing about it!

delvedor:
> Should we throw an error if the user specify address  or port with socket?

Mh, I'm not sure, probably is enough to document the order of our checks.
`address > socket > port`.

allevo:
I don't know why travis fails. Restarting the travis job fixes the problem

PR #19, opened 2017-10-19
Avoid to rewrite the error

allevo:


examples/package-not-found.js
```diff
@@ -0,0 +1,3 @@
+'use strict'
```
delvedor: I'll put this files inside a `test` folder :)

cli.js
```diff
@@ -28,30 +28,33 @@ function start (opts) {
...
 }
 
 function runFastify (opts) {
   var file = null
   try {
     file = require(path.resolve(process.cwd(), opts._[0]))
   } catch (e) {
-    console.log(`Cannot find the specified file: '${opts._[0]}'`)
-    stop(1)
+    return module.exports.stop(e)
```
delvedor: Why?

examples/package-not-found.js
allevo: All tests are inside examples (`incorrect-plugin.js` too)

cli.js
allevo: Because I can ovewrite it during tests

delvedor: I don't like so much as solution, is not elegant. But for the moment could be ok. :)

allevo: Me too. But I have not sadly found another solution!

Issue #16, opened 2017-10-13
support for async-await plugins


mcollina:
As titled, fastify now supports them!

dottorblaster:
Hi!

Do you mean something like this as an async/await plugin?

```js
'use strict'

module.exports = function (fastify, options, next) {
  fastify.get('/', async function (req, reply) {
    var res = await new Promise(function (resolve) {
      setTimeout(resolve, 200, { hello: 'world' })
    })
    return res
  })
  next()
}
```

delvedor:
Hi @dottorblaster!
Nope, he means that the plugin can use async/await, so:
```js
'use strict'

module.exports = async function (fastify, options) {
  fastify.get('/', async function (req, reply) {
    return { hello: 'world' }
  })
}
```

Checkout [avvio](https://github.com/mcollina/avvio) if you want to see how we handle plugins :)

dottorblaster:
Nice! Giving this a shot as soon as possible 😉 Thanks for the support!

mcollina:
Done

dottorblaster:
![Perfect!](https://media.giphy.com/media/Rr3lgUreqh5V6/giphy.gif)

giovanni-bertoncelli:
I haven't found anything in the docs: an async plugin should not use the `done` parameter then? It has only to wait that the promise resolves, right?

alemagio:
Exactly.
You can see an example [here](https://github.com/fastify/fastify-cli/blob/master/templates/app/plugins/support.js)

PR #20, opened 2017-10-26
Add async/await plugin support

dottorblaster:
Addressing #16, I added a test to check the CLI against an async/await plugin, then I added the related example both in README and in the proper directory. I used fastify 0.30.2 to ensure I was working with the very last version of the framework, is it ok?

I had to rework the arity check too, I dont know if it's ok for you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@ You can easily run it with:
 ```bash
 $ fastify plugin.js
 ```
+You can use `async` functions too, and make your plugin more concise: 
+```js
+// async-await-plugin.js
+module.exports = async function (fastify, options) {
+  fastify.get('/', async function (req, reply) {
+    return { hello: 'world' }
+  })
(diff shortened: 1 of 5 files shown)


dottorblaster:
Ouch, Node 4/5/6 is not supporting async/await so CI fails dramatically.  What do we do?

delvedor:
> Ouch, Node 4/5/6 is not supporting async/await so CI fails dramatically. What do we do?

Usually we do something like the following:
```js
test('async await support', t => {
  if (Number(process.versions.node[0]) >= 8) {
    // run the test
  } else {
    tap.pass('Skip because Node version < 8')
    tap.end()
  }
})
```

dottorblaster:
Fixed the test implementation to skip older Node versions and rebased to keep the history clean. Now we are good to go 🎉

PR #24, opened 2017-12-17
Move fastify from devDependency to dependency.

coreyfarrell:
This causes fastify to be installed as a dependency of fastify-cli, as
fastify will no longer provide a command-line utility.

See fastify/fastify#514 for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Write and run a route with one single command!
 
 ## Install
 ```bash
-npm install fastify-cli --save
+npm install fastify-cli --global
 ```
 
 ## Usage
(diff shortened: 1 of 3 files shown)


mcollina:
I think we should verify which version of fastify the plugin is depending on, if any.
It should match with the version the CLI is using, and also check which version is available on npm.

coreyfarrell:
Sorry I don't understand your comment.  What plugin?

mcollina:
With fastify-cli, you can:

fastify myapp.js

myapp.js has a package.json file, and it might have a fastify dependency. the local version should match the one used in the cli, e.g. they should be in sync, or we should show a warning. As an example, gulp does the same.

coreyfarrell:
My new patch causes cli.js to import the best copy of fastify.  If possible we use the copy installed with myapp.js but fastify-cli bundles a copy as a fallback.

PR #27, opened 2018-01-04
Fix #17

### LINKED ISSUE #17: since v0.30 log level is not fatal anymore
Logging is disabled by default in fastify, however I think in the CLI we should default it to info, and maybe throw in `pino-colada` as well (maybe), and __definitely remove the `console.log` statements__.

cc @allevo @delvedor

allevo:


allevo:
@mcollina  `npm install` fails on node 4. Job restarted

allevo:
@mcollina I'm not able to merge on this repo. Anyway if you want to wait for @delvedor it's ok for me. I'm not in hurry.

cli.js
```diff
@@ -120,7 +120,8 @@ function runFastify (opts) {
     if (typeof address === 'object') {
       address = `http://localhost:${fastify.server.address().port}`
     }
-    console.log(`Server listening on ${address}`)
+    fastify.log.info(`Server listening on ${address}`)
+    console.log()
```
mcollina: why are we keeping an empty `console.log` here?

cli.js
```diff
@@ -120,7 +120,8 @@ function runFastify (opts) {
     if (typeof address === 'object') {
       address = `http://localhost:${fastify.server.address().port}`
     }
-    console.log(`Server listening on ${address}`)
+    fastify.log.info(`Server listening on ${address}`)
```
mcollina: should we move this log line to fastify itself?

allevo: It's ok for me

mcollina:
@allevo can you send a PR to fastify to add that logging statement?

PR #34, opened 2018-01-16
Add generate-readme script

allevo:
Add script to keep the README updated with `help.txt` file

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "lint": "standard",
     "unit": "tap test.js",
     "test": "npm run lint && npm run unit",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "lint": "standard",
     "unit": "tap test.js",
-    "test": "npm run lint && npm run unit"
+    "test": "npm run lint && npm run unit",
+    "generate-readme": "sed -e '/\\${help.txt}/r help.txt' -e '/\\${help.txt}/d' README.tmpl  > README.md"
```
mcollina: Can you use node for this instead of sed? I would like to keep the env windows-friendly.

allevo: Updated

Issue #43, opened 2018-03-09
Support bodyLimit from cli


daaaaa:
As titled, allow --bodyLimit to be accepted from fastify-cli

allevo:
Hi daaaaa, thank you for checking out this project!

I'll add the options as soon as possible.

delvedor:
@daaaaa PRs are welcome

daaaaa:
Well, you're right. I'll try to to it during this week.

PR #49, opened 2018-03-18
Add bodyLimit support through --body-limit option

daaaaa:
As requested by myself in #43, I added support for [`bodyLimit` parameter](https://github.com/fastify/fastify/blob/master/docs/Factory.md#bodylimit) passed to the fastify factory through the `--body-limit` option. No default value is set, since fastify has already its default value.

I added the keyword `body-limit` to the `minimist` options in the `integer` array, for coherency, even though `minimist` does not do anything at all with the `integer` key (as we can see from [its documentation](https://github.com/substack/minimist#var-argv--parseargsargs-opts))
Does it have any sense for the `integer` key to exists in the `minimist` options?

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,9 @@ Usage: fastify [opts] <file>
   -h, --help
       Show this help message
 
+  --body-limit
+      Defines the maximum payload, in bytes, the server is allowed to accept
+
 ```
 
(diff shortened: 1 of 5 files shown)


mcollina:
You might want to add a unit test for this.

mcollina:
Thanks!

PR #53, opened 2018-04-03
Support also env vars as options

delvedor:
It could happen that a user can't pass the options via cli args, such as when using docker and dynamic arguments.

With this pr we add support for the same parameters also from environment variables. Every variable is prefixed with `FASTIFY` to avoid conflicts with other configurations. 

cc @ninjatux


delvedor:
I've also added [`dotenv`](https://www.npmjs.com/package/dotenv).
To test it I had to add an `.env` file in the root directory.

ninjatux:
seems good to me! 👍

README.md
```diff
@@ -90,6 +90,21 @@ module.exports.options = {
...
+| ------------- | ------------- |-------------| ----- |
+| Port to listen on (default to 3000) | `-p` | `--port` | `FASTIFY_PORT` |
+| Address to listen on | `-a` | `--address` | `FASTIFY_ADDRESS` |
+| Socket to listen on | `-s` | `--socket` | `FASTIFY_SOCKET` |
+| Log level (default to fatal) | `-l` | `--log-level` | `FASTIFY_LOG_LEVEL` |
+| Prints pretty logs | `-P` | `--pretty-logs` | `FASTIFY_PRETTY_LOGS` |
+| Use custom options | `-o` | `--options` | `FASTIFY_OPTIONS` |
+| Set the prefix | `-p` | `--prefix` | `FASTIFY_PREFIX` |
+| Defines the maximum payload, in bytes,<br>the server is allowed to accept |  | `--body-limit` | `FASTIFY_BODY_LIMIT` |
+
```
mcollina: You need to mention dotenv here as well!

Issue #57, opened 2018-04-09
Bind 0.0.0.0 in docker env


delvedor:
If you run a Fastify server in Docker by default it will run on the address `127.0.0.1` which is not reachable from outside.
We can add an [*is docker*](https://github.com/sindresorhus/is-docker) check and if it is true, use `0.0.0.0` instead of `127.0.0.1`.

What do you think?

ninjatux:
it's a sensible thing to do if by default we want a working service limiting options and arguments when running `in` docker

mcollina:
I would do this by default if bundled within docker.

pyros2097:
Damn'it I'm too late. Just ran into it when using docker. Would have been easier if I had seen this earlier.

mercteil:
In case someone landed here -> the merged fix did not help me... `server.listen(port, '0.0.0.0')` did

PierBover:
I think something is broken here.

```
fastify.listen(process.env.PORT || 3000, '0.0.0.0', function (err, address) {
	if (err) {
		fastify.log.error(err)
		process.exit(1)
	}
	fastify.log.info(`server listening on ${address}`)
});
```
When running on regular Node I get `Server listening on http://0.0.0.0:3000` as expected.

But when running in Docker I get `Server listening at http://127.0.0.1:8080` so the server cannot be reached.

Wasn't this PR by @mcollina supposed to add `0.0.0.0` by default when running in Docker?

https://github.com/fastify/fastify-cli/pull/58

It seems now it's actually overwriting the address and using `127.0.0.1` instead.

Or maybe I'm missing something...

mcollina:
I do not think you are running your server using fastify-cli. Are you using this module at all?

The snippet above is not from this module.

PierBover:
Ooops sorry @mcollina I totally missed this wasn't the main Fastify repo.

PR #58, opened 2018-04-10
added automatic bind to 0.0.0.0 when running in docker

ninjatux:
this pr solves #57


package.json
```diff
@@ -57,10 +57,13 @@
   },
   "devDependencies": {
     "fastify-autoload": "^0.2.1",
+    "is-docker": "^1.1.0",
```
delvedor: This should be inside the `dependencies` field.

PR #61, opened 2018-04-21
Set the default logger level to INFO in the template

mcollina:
as titled.


delvedor:
Currently is `fatal`.
https://github.com/fastify/fastify-cli/blob/e1d5e2f9de3690df7affcf59a66b8f205518e9f8/start.js#L75

mcollina:
The reason why I wanted INFO is that it prints the port the server it is listening to.

mcollina:
@delvedor what do you think about this? I think it will be highly useful.

PR #62, opened 2018-04-21
Pass options down to the autoloaded plugins

mcollina:
I'm not super convinced about this. Maybe we should rename `dir` to `autoload: { dir }` to avoid namespace clashing.

What do you think?


mcollina:
This does not work well with unit tests. I would like to recycle the db connection, and I can’t pass option through.

mcollina:
let's wait to land this. I'm filing many more PRs for this, as to properly fix one issue I'm going several level deep (down to https://github.com/mcollina/avvio/pull/62).

mcollina:
This is now ready to go.

PR #70, opened 2018-05-11
fix: make generate work with yarn

### LINKED ISSUE #69: Generating project does not work with yarn
Hello!

Generating a new project with yarn does not work, because yarn does not include a scripts property to the package.json.

```
pkg.scripts.test = 'standard && tap test/*.test.js test/*/*.test.js test/*/*/*.test.js'
                     ^
TypeError: Cannot set property 'test' of undefined
```

works fine with npm

tiswrt:
closes #69

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -30,10 +30,12 @@ function generate (dir, log, cb) {
       return cb(err)
     }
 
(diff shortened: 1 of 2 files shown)


tiswrt:
not sure about this approach, let me know what you think :)

PR #80, opened 2018-06-11
fix (generate): properly set 'fastify-plugin'

### LINKED ISSUE #79: lack dependency
the generated `plugins/support.js` file require `fastify-plugin`, but it is not included in the dependencies of package.json

zhaoyao91:
resolve #79


mcollina:
Can you please update the tests as well?

zhaoyao91:
updated

Issue #87, opened 2018-07-16
start: default to process.env.PORT if present


mcollina:
Seems a good thing to add!

delvedor:
Should take precedence over `FASTIFY_PORT` or the args?

mcollina:
It should be at the lowest precedence level.

mcollina:
This was implemented in #88.

PR #88, opened 2018-07-17
added PORT env var

ovhemert:
Added support for PORT env variable when present.
Issue #87


ovhemert:
Help also updated.. missed that one.
Simple, but useful feature to support PORT env

Issue #91, opened 2018-07-28
how to reload file when file was edtied?


gaogao1030:
The purpose in dev is useful,  could add -w options to watch files like nodemon.

mcollina:
Would you like to send a PR? It’d be very useful indeed!

gaogao1030:
@mcollina https://github.com/fastify/fastify-cli/pull/92.

mcollina:
Fixed in #92

PR #92, opened 2018-07-30
added watch env var (#91)

gaogao1030:
* added watch env var

* change test/start.js

* updated help text

### DIFF
--- a/fork.js
+++ b/fork.js
@@ -1,0 +1,7 @@
+const { runFastify } = require('./start')
+
+const fastify = runFastify(process.argv.splice(2))
+
+fastify.ready(err => {
+  process.send({ type: 'ready', err: err })
+})
--- a/fork.js
+++ b/fork.js
@@ -1,7 +1,0 @@
-const { runFastify } = require('./start')
-
-const fastify = runFastify(process.argv.splice(2))
-
-fastify.ready(err => {
-  process.send({ type: 'ready', err: err })
-})
--- a/help/start.txt
+++ b/help/start.txt
@@ -24,6 +24,10 @@ Usage: fastify start [opts] <file>
   [env: FASTIFY_OPTIONS]
       Use custom options
 
+  -w, --watch
+  [env: FASTIFY_WATCH]
+      Watch process.cwd() directory for changes, recursively; when that happens, the process will auto reload.
+
   -r, --prefix
   [env: FASTIFY_PREFIX]
       Set the prefix
--- a/lib/watch/fork.js
+++ b/lib/watch/fork.js
@@ -1,0 +1,31 @@
+const { stop, runFastify } = require('../../start')
+
+const GRACEFUL_SHUT = 'GRACEFUL SHUTDOWN'
+const FULLY_CLOSED = 'FULLY CLOSED'
+
+const fastify = runFastify(process.argv.splice(2))
+const { type } = process.env.childEvent
+let timeout = null
+
+process.send({ type: type, err: null })
+
+fastify.ready(err => {
+  if (err) { stop(err) }
+  process.send({ type: 'ready', err: err })
+})
+
+fastify.addHook('onClose', (instance, done) => {
+  clearTimeout(timeout)
+  process.send({ type: FULLY_CLOSED, err: null })
+  done()
+})
+
+process.on('message', function (event) {
+  if (event === GRACEFUL_SHUT) {
+    fastify.close()
+
+    timeout = setTimeout(function () {
+      process.send({ type: FULLY_CLOSED, err: null })
+    }, 5000)
+  }
+})
--- a/lib/watch/fork.js
+++ b/lib/watch/fork.js
@@ -1,8 +1,11 @@
 const chalk = require('chalk')
 const { stop, runFastify } = require('../../start')
 
-const GRACEFUL_SHUT = 'GRACEFUL SHUTDOWN'
-const READY = 'ready'
+const {
+  GRACEFUL_SHUT,
+  READY,
+  TIMEOUT
+} = require('./constants.js')
 
 const fastify = runFastify(process.argv.splice(2))
 const type = process.env.childEvent
@@ -16,20 +19,23 @@ fastify.ready(err => {
 
 process.on('message', function (event) {
   if (event === GRACEFUL_SHUT) {
-    const timeout = setTimeout(function () {
-      process.exit(1)
-    }, 5000)
-
+    const message = chalk.red('[fastify-cli] process forced end')
+    setTimeout(exit.bind({ message }), TIMEOUT).unref()
     fastify.close(() => {
-      timeout.unref()
       process.exit(0)
     })
   }
 })
 
 process.on('uncaughtException', (err) => {
-  fastify.close()
   console.log(err)
-  console.log(chalk.red('[fastify-cli] app crashed - waiting for file changes before starting...'))
-  process.exit(1)
+  const message = chalk.red('[fastify-cli] app crashed - waiting for file changes before starting...')
+  // fastify.close(exit) // It's looks that only happend fastify normally startup. so we should detect fastify is normally runing, it's normally running that graceful exit happen, other case I think it's better immediately exec process.exit(1)
+  // setTimeout(exit.bind({ message }), TIMEOUT).unref()
+  exit.bind({ message })()
 })
+
+function exit () {
+  if (this) { console.log(this.message) }
+  process.exit(1)
+}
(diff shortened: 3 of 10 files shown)


help/start.txt
```diff
@@ -26,7 +26,7 @@ Usage: fastify start [opts] <file>
 
   -w, --watch
   [env: FASTIFY_WATCH]
-      Watch files that the file is change, the process will auto reload
+      Watch process.cwd()'s directory that the directory is change, the process will auto reload
```
mcollina: Watch `process.cwd()` directory for changes, recursively; when that happens, the process will auto reload.

gaogao1030: Thank you very much.

gaogao1030:
@mcollina can you review about watch functionality code and test, my machine test is all ok. but ci always failed https://travis-ci.org/gaogao1030/fastify-cli/jobs/411331476. i don't know how to solve this failed. i need some help.

test/start.js
```diff
@@ -433,3 +401,38 @@ test('should start the server listening on 0.0.0.0 when runing in docker', t =>
...
+      if (fs.existsSync(tmpjs)) {
+        fs.unlinkSync(tmpjs)
+      }
+    })
+
+    fastifyEmitter.on('start', () => {
+      t.pass('should receive start event')
+      setTimeout(function () {
+        t.pass('touch tmpjs')
+        touch.sync(tmpjs)
```
gaogao1030: it's looks like `touch.sync(tmpjs)` no trigger file change event or that event no caught by chokidar watch in Travis-CI.

fork.js
```diff
@@ -0,0 +1,7 @@
+const { runFastify } = require('./start')
+
+const fastify = runFastify(process.argv.splice(2))
+
+fastify.ready(err => {
+  process.send({ type: 'ready', err: err })
+})
```
mcollina: You might want to receive a signal about the parent a) closing and b) wanting to shut down this. Ideally we should shut fastify down gracefully.

gaogao1030:
I am updated gracefully shutdown commit but CI still build faild, it's looks chokidar.watch can't caught file change event when file changed.

lib/watch/fork.js
```diff
@@ -0,0 +1,31 @@
...
+
+fastify.addHook('onClose', (instance, done) => {
+  clearTimeout(timeout)
+  process.send({ type: FULLY_CLOSED, err: null })
+  done()
+})
+
+process.on('message', function (event) {
+  if (event === GRACEFUL_SHUT) {
+    fastify.close()
```
mcollina: You can pass a callback to `fastify.close()`, and avoid the ` addHook` call.

lib/watch/fork.js
```diff
@@ -0,0 +1,31 @@
...
+
+    timeout = setTimeout(function () {
+      process.send({ type: FULLY_CLOSED, err: null })
+    }, 5000)
```
mcollina: can you `unref()`  the timer? I would go with "process.exit(1)" instead.

test/start.js
```diff
@@ -433,3 +402,62 @@ test('should start the server listening on 0.0.0.0 when runing in docker', t =>
...
+    })
+
+    fastifyEmitter.on('restart', () => {
+      t.pass('should receive restart event')
+    })
+
+    fastifyEmitter.on('ready', () => {
+      t.pass('should receive ready event')
+    })
+  }, 500)
```
mcollina: Can you remove the various setTimeout? It seems that chokidar itself has some problem with its CI. We can skip these tests on Travis I presume. What do you think @delvedor?

test/start.js
```diff
@@ -433,3 +402,62 @@ test('should start the server listening on 0.0.0.0 when runing in docker', t =>
...
+      fastifyEmitter.emit('close')
+    })
+
+    fastifyEmitter.on('start', () => {
+      t.pass('should receive start event')
+      setTimeout(function () {
+        t.pass('touch tmpjs')
+        // fastifyEmitter.emit('ready')
+        // fastifyEmitter.emit('restart')
+        exec(`touch ${tmpjs}`) // chokidar watch can't caught change event in travis CI, but local test is all ok. you can remove annotation in local environment.
```
mcollina: Can you avoid using bash? I would prefer the tests to be able to run on Windows as well.

test/start.js
```diff
@@ -433,3 +402,62 @@ test('should start the server listening on 0.0.0.0 when runing in docker', t =>
...
   })
 })
+
+test('should start the server with watch options that the child process restart when directory changed', t => {
+  t.plan(5)
+
+  const tmpjs = path.resolve(baseFilename + '.js')
+  fs.writeFileSync(tmpjs, 'hello world')
+
+  setTimeout(function () {
```
mcollina: We should be able to have this pass without these massive setTimeout.

lib/watch/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+      if (type === FULLY_CLOSED) {
+        childs.shift().kill('SIGUSR2')
+        return null
+      }
+
+      if (type === READY) {
+        setTimeout(function () {
+          emitter.emit(type)
+        }, 100)
```
mcollina: Why is this `setTimeout` needed?

lib/watch/index.js
```diff
@@ -0,0 +1,86 @@
...
+      const { type, err } = childEvent
+      if (err) {
+        emitter.emit('error', err)
+        return null
+      }
+
+      if (type === FULLY_CLOSED) {
+        childs.shift().kill('SIGUSR2')
+        return null
+      }
```
mcollina: why do we need to wait for this signal?

gaogao1030: Only use SIGUSR2 signal kill child process, the main process would fork new child process. other signals don't.

mcollina: I mean, why are waiting for a FULLY_CLOSED message from the child process before restarting?

gaogao1030: Because the signal would send in fastify.close() callback or hook timing that call in child_process.

gaogao1030: should child_process killed in itself instead of don't send message parent process?

gaogao1030: Because new child process should fork in fastify.close() timing and old child process would kill.

mcollina: I think we should just `process.exit(0)`  in the child process within the `fastify.close()` callback. You should not be waiting for a message.

lib/watch/fork.js
```diff
@@ -0,0 +1,35 @@
...
+})
+
+process.on('message', function (event) {
+  if (event === GRACEFUL_SHUT) {
+    const timeout = setTimeout(function () {
+      process.exit(1)
+    }, 5000)
+
+    fastify.close(() => {
+      timeout.unref()
```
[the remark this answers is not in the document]
gaogao1030: It fastify has provided the hook function of before .close() ? I think it's unref() should call fastify.close() before or fastify.close() callback。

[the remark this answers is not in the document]
mcollina: Calling `unref()` before `process.exit(0)` is useless (we will exit anyway). If for some reason an application is not calling the `onClose()` callbacks, this will hang for 5s. calling `unref()` would make it possible for that application to exit normally.

lib/watch/fork.js
```diff
@@ -0,0 +1,35 @@
...
+      process.exit(0)
+    })
+  }
+})
+
+process.on('uncaughtException', (err) => {
+  fastify.close()
+  console.log(err)
+  console.log(chalk.red('[fastify-cli] app crashed - waiting for file changes before starting...'))
+  process.exit(1)
```
mcollina: we should have:

```js
process.on('uncaughtException', (err) => {
  console.log(err)
  fastify.close(exit) 
  setTimeout(exit, 5000).unref() // so that we wait only 5s for a graceful exit.

function exit () {
    console.log(chalk.red('[fastify-cli] app crashed - waiting for file changes before starting...'))
      process.exit(1)
  })
 
})
```

lib/watch/fork.js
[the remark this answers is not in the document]
gaogao1030: thank you very much. I get it.

lib/watch/fork.js
gaogao1030: `fastify.close(exit)` it's looks that only happend fastify normally startup. so we should detect fastify is normally runing, it's normally running that graceful exit happen, other case I think it's better immediately exec process.exit(1)

gaogao1030: could we should remove `fastify.close(exit)` and ` setTimeout(exit, 5000).unref()`  ?  it's  graceful exit looks never happend in uncaughtException. I think it's better immediately exec process.exit(1).

Issue #94, opened 2018-08-24
Consider removing linter tasks from the generated project


cemremengu:
Currently the cli generates a project with the `standardjs` linting. However, most people is probably using eslint (I suppose). So I believe that the cli should support one of the following:

a) A generator option with eslint config  (maybe just use airbnb preset)
b) A bare and unopinionated generator in terms of linting (Simpler)

mcollina:
I do not think we should add options. The moment we start adding options we will get:

- choice of linter
- choice of testing framework
- choice between callbacks and async/await
- choice of database
- choice of template engine
- ...

The generator _is_ opinionated. It is how @delvedor and myself (and possibly others from @fastify/fastify) think an application should be structured. You can structure your application the way you prefer (even all the code in one file).

What do you mean by unopinionated? Should we just remove the `linter` task?

cemremengu:
Yes, you are right. One of the things I like about fastify that it is not bloated so definitely not a fan of the choice tree you mentioned. 

By unopinionated, I only mean the linting choice of  `standardjs` in the generated project. I can imagine many people editing/changing `standardjs` parts in `devDependecies` and `scripts` after creating the project. 

Since linting is purely based on choice at the moment in JS ecosystem, it may be a good idea to describe linting task integration in readme instead of forcing it to the cli.

mcollina:
This is interesting, ok!

@delvedor what do you think? It might be good to do!

delvedor:
I agree that directly add `standard` as a dependency is opinionated, but if we make it optional, what about the generated code? It follows standard's rules.

mcollina:
I would keep the code as it is right now.
Il giorno ven 24 ago 2018 alle 14:29 Tomas Della Vedova <
notifications@github.com> ha scritto:

cemremengu:
I think having the code following standard's rules should be fine, formatters/linters can change them automatically. 

After all, you need to be opinionated on the generated code based on the rule set of choice. There seems to be no other way.

allevo:
I'm 👎 to remove.
`standard` as a dependency is opinionated as the generated project structure is.
If `generate` command reflects what we (???) suggests, the linting script should be included too.

Anyway I'm not blocking

jsumners:
I think that if you have strong opinions about project structure, which includes the linter used, then a third party code generator is not for you.

mcollina:
[...]

PR #95, opened 2018-08-24
Remove linter task from the generated code

cemremengu:
As discussed in #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,6 @@ The sample code offers you four npm tasks:
   [`pino-colada`](https://github.com/lrlna/pino-colada) pretty logging
   (not suitable for production)
 * `npm test` - runs the tests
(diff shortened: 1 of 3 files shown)


mcollina:
the test would need to be updated: https://github.com/fastify/fastify-cli/blob/master/test/generate.js#L119

README.md
```diff
@@ -142,6 +141,24 @@ You will find three different folders:
 Finally there will be an `app.js` file, which is your entry point.
 It is a standard Fastify plugin and you will not need to add the `listen` method to run the server, just run it with one of the scripts above.
 
+### linting
+
+`fastify-cli` is unopinionated on the choice of linter. You can integrate `Standard` as follows:
```
mcollina: can you change this to

> We recommend you to add a linter, like so:

PR #98, opened 2018-08-28
Unit tests are passing again.

mcollina:
I do not know why they failed so badly. I've also noticed that they were not really reliable as they where written. I refactored the whole lot and now are passing again on my machine.


test/start.js
```diff
@@ -14,30 +14,34 @@ const start = require('../start')
 
 const onTravis = !!process.env.TRAVIS
 
+let _port = 3001
+
+function getPort () {
```
delvedor: Why not generate a random port?

mcollina: This worked, random would be trickier.

PR #105, opened 2018-09-03
Make sure to restart after change on app crash

### LINKED ISSUE #96: Watch does not restart the process once it crashes
Message says the process is crashed and waiting for changes to restart however nothing happens after changes. Seeing the comments here, is this a WIP feature? I am using node v8.11.3 on windows 10.

```
process.on("uncaughtException", err => {
  console.log(err);
  const message = chalk.red(
    "[fastify-cli] app crashed - waiting for file changes before starting..."
  );
  // fastify.close(exit) // It's looks that only happend fastify normally startup. so we should detect fastify is normally runing, it's normally running that graceful exit happen, other case I think it's better immediately exec process.exit(1)
  //setTimeout(exit.bind({ message }), TIMEOUT).unref();
  exit.bind({ message })();
});
```

cemremengu:
Fixes #96

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -75,6 +75,7 @@ const watch = function (args) {
           console.log(err)
           stop(watcher, err)
         }
+        childs.push(run('restart'))
       }
     })
   })


mcollina:
How can I test this?

cemremengu:
I think I have an idea

cemremengu:
Nevermind, I talked too early. I think chokidar watch can't catch change events in travis CI anyways 😞

lib/watch/index.js
```diff
@@ -75,6 +75,7 @@ const watch = function (args) {
           console.log(err)
           stop(watcher, err)
         }
+        childs.push(run('restart'))
```
mcollina: I think this whole try-catch needs a fix. The condition `if (!childs.length === 0)`  can never be `true`, because `!childs.length`  will either be `true`  or `false`, but never `0`.

cemremengu: I think it actually meant length !== 0 which I changed in #107

PR #107, opened 2018-09-03
Cleanup

cemremengu:
Removed comments and refactored some lines

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -15,11 +15,10 @@ let allStop = false
 let childs = []
 
 const stop = (watcher = null, err = null) => {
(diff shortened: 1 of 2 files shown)


lib/watch/index.js
```diff
@@ -15,11 +15,8 @@ let allStop = false
 let childs = []
 
 const stop = (watcher = null, err = null) => {
-  if (childs.length > 0) {
-    childs.forEach(function (child) {
-      child.kill()
-    })
-  }
+  childs.map(child => child.kill())
```
mcollina: why are you using `map` here?

PR #109, opened 2018-09-09
Fix docs on how to migrating out of fastify-cli start (#104)

x1yy:
When i follow the instruction i get error:
```
ReferenceError: fastify is not defined
```

The right reference is the `app`, isn't it?

Here is the working snippet:
```
// Register your application as a normal plugin.
app.register(require('./app.js'))

// Start listening.
app.listen(process.env.PORT || 3000, (err) => {
  if (err) {
    app.log.error(err)
    process.exit(1)
  }
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,12 +137,12 @@ const app = Fastify({
 })
 
 // Register your application as a normal plugin.
-fastify.register(require('./app.js'))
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks!

Issue #110, opened 2018-09-28
some env variables doesn't work


clarkorz:
Three env variables `FASTIFY_PRETTY_LOGS`, `FASTIFY_OPTIONS`, `FASTIFY_WATCH`  doesn't work. they will be override by default values.

[Related code](https://github.com/fastify/fastify-cli/blob/master/start.js#L103):
```
// run without any command line arg
// $ fastify start app.js
let opts = minimistArgs(args)  // --> contain { pretty-log: false, options: false, watch: fasle }
opts = Object.assign(readEnv(), opts) -> opts will override env
```

mcollina:
Would you like to send a PR to fix this?

Il giorno ven 28 set 2018 alle 12:38 Clark Wang <notifications@github.com>
ha scritto:

PR #111, opened 2018-10-02
Fix boolean env variables override

JGiola:
Swapped minimist with yargs-parser because works better with prefixed env variables and is currently maintained.

This move will fix issue #110.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -7,7 +7,7 @@ const {
 const chalk = require('chalk')
 const path = require('path')
 const generify = require('generify')
-const minimist = require('minimist')
+const argv = require('yargs-parser')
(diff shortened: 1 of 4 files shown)


cemremengu:
How about using [meow](https://github.com/sindresorhus/meow) ?

JGiola:
I'am sure it will work because it has the yargs-parser as a dependency. I've selected because it's used in [yargs, that is the library indicated by the developer of optimist as its successor, and is the parser used by the nyc and angular-cli project.

cemremengu:
@JGiola fair enough, thanks for the info!

cemremengu:
@delvedor ping

PR #113, opened 2018-10-16
Testing windows on travis

### LINKED ISSUE #112: Tests are failing on windows
Most tests are failing for me due to comparison errors like

```js
test/generate.js .................................... 47/67
  finish succesfully if package.json is there - npm
  not ok \app.js matching
    +++ found
    --- wanted



                      require('fastify-autoload')

                 ts = function (fastify, opts, next) {
                                 code!

                             llowing lines

                     all plugins defined in plugins
              e should be support plugins that are reused


                         _dirname, 'plugins'),
                              n({}, opts)


                   s all plugins defined in services
                           s in one of these

                        __dirname, 'services'),
                              n({}, opts)


                             ext when done


    +'use strict'
    +
    +const path = require('path')
    +const AutoLoad = require('fastify-autoload')
    +
    +module.exports = function (fastify, opts, next) {
    +  // Place here your custom code!
    +
    +  // Do not touch the following lines
    +
    +  // This loads all plugins defined in plugins
    +  // those should be support plugins that are reused
    +  // through your application
    +  fastify.register(AutoLoad, {
    +    dir: path.join(__dirname, 'plugins'),
    +    options: Object.assign({}, opts)
    +  })
    +
    +  // This loads all plugins defined in services
    +  // define your routes in one of these
    +  fastify.register(AutoLoad, {
    +    dir: path.join(__dirname, 'services'),
    +    options: Object.assign({}, opts)
    +  })
    +
    +  // Make sure to call next when done
    +  next()
    +}
```

The expected stuff is coming out garbled. I am suspecting `\r\n` situation but I couldn't figure out the reason so shamelessly posting here. 😄 

Any ideas?

cemremengu:
Fixes #112 

cc @mcollina looks like `CRLF/LF`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: node_js
+
+os:
+  - windows
+  - linux
+  
 node_js:
   - "8"
   - "6"
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "dotenv": "^6.0.0",
     "fastify": "^1.11.0",
     "fastify-plugin": "^1.2.0",
-    "generify": "^3.0.0",
+    "generify": "^3.0.2",
     "help-me": "^1.1.0",
     "is-docker": "^1.1.0",
     "make-promises-safe": "^1.1.0",
--- a/test/start.js
+++ b/test/start.js
@@ -84,7 +84,7 @@ test('should start the server at the given prefix', t => {
   })
 })
 
-test('should start fastify at given socket path', t => {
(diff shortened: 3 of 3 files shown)


cemremengu:
I think we need to update random port selection to be > 1024 if on windows?

mcollina:
I would just skip https://travis-ci.org/fastify/fastify-cli/jobs/442283718#L186. Unix sockets are not well  supported on windows.

cemremengu:
@mcollina cant merge this one

Issue #89, opened 2018-07-25
Getting started with a new project


benhalverson:
Expected 
`fastify generate` creates a new project as the docs state.

Actual result
```
--> reading package.json in /Users/benhalverson/projects/fastify-cli-test
ENOENT: no such file or directory, open '/Users/benhalverson/projects/fastify-cli-test/package.json'
```

The workaround is to create a package.json file first in directory your project will be in
`npm init -y`

delvedor:
As [documented](https://github.com/fastify/fastify-cli#generate) `fastify generate` expects that the `package.json` is already created, so this is not an issue.

@mcollina should we think about create the `package.json` is is not present?
In which case probably we should also create the directory.

cemremengu:
Landed in #114 (v0.25.0)

PR #114, opened 2018-10-18
Project scaffolding with package.json

cemremengu:
This aims to address #89 and in general achieve a better generator

The tests are failing intentionally now since I am keen on introducing a breaking change about existing `package.json` file 😄 

In my opinion, the `generate` command should error if there is already an existing `package.json`. This both simplifies things, prevents overwriting mistakenly, enables more control on `package.json` (with different templates) and, in general, I think it is what most generators  do and people expect.

Once we agree on the approach, I will adjust tests & final code accordingly.

Questions for bindings:
- project name
- project description
- author

New behavior:
- `generate my-dir`  creates a new project inside a new directory with the contents of `app_template` bound
- `generate` creates a new project with `package.json` in the current directory with the contents of `app_template` bound
- A new directory is only created if it does not exists
- If a `package.json` file exists in the target directory project generation fails


question.js
```diff
@@ -0,0 +1,21 @@
...
+    },
+    {
+      type: 'input',
+      name: 'description',
+      message: 'project description'
+    },
+    {
+      type: 'input',
+      name: 'author',
+      message: 'project author'
```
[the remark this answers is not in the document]
cemremengu: I did this but the template has `app.j` as the entrypoint. Is it a problem to change it to `index.js` ?

Good thing about asking our own questions was that we can alter the process as needed.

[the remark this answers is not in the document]
mcollina: We are already editing the package.json file, I’m good in changing it post-fact. I’m also ok in asking our own question, as long as we are consistent with npm init configs.

question.js
```diff
@@ -0,0 +1,63 @@
...
+          return 'Project name may only include letters, numbers, underscores and hashes.'
+        }
+      }
+    },
+    {
+      type: 'input',
+      name: 'version',
+      message: 'version',
+      validate: function (input) {
+        if (/[0-9]+\.[0-9]+\.[0-9]+/.test(input)) {
```
mcollina: I think we should be using the semver module to validate this.

cemremengu:
`Where is this loading the data from the npm config?`

Sorry could you please elaborate? I decided that asking custom questions is better so it is not using `npm init` if that's what you are wondering.

`npm init` asks questions that we modify later on such as `entry point` and `scripts` so it can frustrate or confuse the user. I checked some other generators and they seem to use custom questions as well.

question.js
cemremengu: Sure will patch in once github status is green

mcollina:
I think we should be reading the actual npm file. How do other generators deal with this? 

https://github.com/npm/cli/blob/latest/lib/config/core.js and https://www.npmjs.com/package/config-chain reads npm config.

cemremengu:
By npm config if you mean `package.json` there is a [template]('https://github.com/fastify/fastify-cli/pull/114/files#diff-81f85c0bd61de03b89eab5c1200d4071) that relies on `generify` templating with the variables from questions.

Instead, we can run `npm init -y` that generates the package file without asking questions and then read & modify the contents. Does that sound better?

mcollina:
> Instead, we can run npm init -y that generates the package file without asking questions and then read & modify the contents. Does that sound better?

+1, definitely better

cemremengu:
@mcollina ping. please review when you get a chance.

cemremengu:
@mcollina do you mind giving me permission to merge?

@delvedor It would be great to have your review as well for this.

mcollina:
You should be able to merge and release now.

cemremengu:
@mcollina published a minor for previous changes but cant bump version on master

> remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "continuous-integration/travis-ci" is expected. At least 1 approving review is required by reviewers with write access.
To https://github.com/fastify/fastify-cli
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/fastify/fastify-cli'

mcollina:
should be fixed now

Il giorno dom 28 ott 2018 alle ore 21:55 Cemre Mengu <
notifications@github.com> ha scritto:

app_template/.gitignore
```diff
@@ -0,0 +1,66 @@
...
+# vscode
+.vscode
+*code-workspace
+
+# clinic
+profile*
+*clinic*
+*flamegraph*
+
+# lock files
```
andriyor: Why file lock files is ignored?

nwoltman: This may have just been copypasta. I've created #188 to remove this.

PR #125, opened 2018-11-14
Simplify testing script

### LINKED ISSUE #124: Default test paths for the template
If Fastify assumes the recursive folder structure, why wouldn't change the default test paths from `tap test/*.test.js test/*/*.test.js test/*/*/*.test.js` to `tap test/**/*.test.js` for the template, because it ignores nested plugin/service folders

cemremengu:
Fixes #124 

Although it does not seem to create a problem (at least for me), it's a nicer way.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -44,7 +44,7 @@ function generate (dir, log, cb) {
       pkg.main = 'app.js'
 
       pkg.scripts = Object.assign(pkg.scripts || {}, {
-        'test': 'tap test/*.test.js test/*/*.test.js test/*/*/*.test.js',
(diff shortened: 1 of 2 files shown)


delvedor:
Hi, sorry for the delay!
That script has been updated in https://github.com/fastify/fastify-cli/pull/51 with the following reason:
*Now the test runner path is `test/*.test.js test/**/*.test.js test/**/**/*.test.js`, because it can happen that a user writes some test in a deeper folder, `/test/services/example/example.test.js` for example.*

We should write a test to verify that case :)

PR #117, opened 2018-10-30
Add fastify-cli.config.js to custom ignored directory in watch mode

gaogao1030:
Ignore configurable for watch mode

PR #126, opened 2018-11-15
Add ignore-watch env variable to ignore changes to the specified file or directory under the process.cwd() directory

gaogao1030:
https://github.com/fastify/fastify-cli/pull/117#issuecomment-434320703

### DIFF
--- a/lib/watch/utils.js
+++ b/lib/watch/utils.js
@@ -1,0 +1,11 @@
+const arrayToRegExp = (arr) => {
+  const reg = arr.map((file) => {
+    if (/^\./.test(file)) { return `\\${file}` }
+    return file
+  }).join('|')
+  return new RegExp(`(${reg})`)
+}
+
+module.exports = {
+  arrayToRegExp
+}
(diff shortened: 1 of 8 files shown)


lib/watch/utils.js
```diff
@@ -0,0 +1,11 @@
+const arrayToRegExp = (arr) => {
```
mcollina: Can you add 'use strict' at the top?

test/watch.test.js
```diff
@@ -0,0 +1,14 @@
+const { arrayToRegExp } = require('../lib/watch/utils')
```
mcollina: use strict at the top

start.js
```diff
@@ -65,7 +65,7 @@ function start (args, cb) {
   })
 
   if (opts['watch']) {
-    return watch(args)
+    return watch(args, parsedArgs(args))
```
cemremengu: This will parse all the options again. You can just use the already parsed `opts` from above.

Also you will need to rebase these on top of #129 :(

Issue #127, opened 2018-11-16
Error when .env set bodyLimit


NemoAlex:
```
TypeError: 'bodyLimit' option must be an integer > 0. Got '5242880'
```
Looks like dotenv converted it to a string.

mcollina:
Ouch, would you like to send a PR?

cemremengu:
@NemoAlex Landed in `v0.25.4`

PR #128, opened 2018-11-16
Fix: wrong parameter with yargs-parser

NemoAlex:
#127

### DIFF
--- a/start.js
+++ b/start.js
@@ -81,7 +81,7 @@ function stop (error) {
 
 function parsedArgs (args) {
   return argv(args, {
-    integer: ['port', 'body-limit'],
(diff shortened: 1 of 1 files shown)


cemremengu:
Do you think you can add a unit test for `parsedArgs` function ?

NemoAlex:
@cemremengu Sorry, recently I don't have time to do that.

cemremengu:
@NemoAlex sure no problem

cemremengu:
@mcollina I will add tests before patching this

cemremengu:
@NemoAlex was this really the solution to your problem? I can't seem to reproduce it. number or integer both works fine for me.

PR #129, opened 2018-11-16
Update arg parsing

cemremengu:
Updates the way args is parsed and used

### DIFF
--- a/start.js
+++ b/start.js
@@ -5,9 +5,7 @@
 const path = require('path')
 const fs = require('fs')
 const assert = require('assert')
-
 const updateNotifier = require('update-notifier')
-const argv = require('yargs-parser')
 const PinoColada = require('pino-colada')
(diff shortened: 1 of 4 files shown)


start.js
```diff
@@ -79,37 +78,10 @@ function stop (error) {
...
-      'options': false
-    }
-  })
-}
-
 function runFastify (args, cb) {
   require('dotenv').config()
-  let opts = parsedArgs(args)
+  let opts = parseArgs(args)
   opts.port = opts.port || process.env.PORT || 3000
```
cemremengu: @mcollina should this be `PORT` or `FASTIFY_PORT` ? I think the latter right?

mcollina: it should be PORT. it’s a well known env variable

cemremengu: Ok, I removed it from README initially so adding it back now

PR #137, opened 2018-11-23
Update cli.js

patrickcurl:
On linux, this is throwing an error --having a space between #! and /usr/bin/env node ... removed the space seems to make it work.


patrickcurl:
Yeah, if I clone, and run node cli.js -- I get the same error, after patching it, node cli.js works fine and generate runs..

PR #144, opened 2018-12-17
Feature/project readme generator

StarpTech:
This is the first draft for a README generator. It should encourage plugin developers to provide a concise plugin readme of a fastify plugin. Consumers can immediately check what decorators a plugin provides, on which dependencies it relies and more ... All the metainformation can be extracted from a fastify plugin.

Some notes about fastify-cli: The implementation of the command was straight but writing tests is no fun e.g the current setup requires that all generated files must exist in the template dirs otherwise they will be deleted on the next run. In general, I got the feeling that the current setup isn't designed to get bigger and bigger. There are really awesome tooling out there to create a maintainable cli e,g (https://github.com/oclif/oclif). Writing callback-style in 2018 is also a bit headache for me. At some place sync and async methods are used without meaningful reason.

It would be awesome if someone could create a test for my command.

### Command
```
fastify generate-plugin ./plugin.js
```

### Input

```js
const fp = require('fastify-plugin')

function plugin (fastify, opts, next) {
  // your plugin code
  next()
}

module.exports = fp(plugin, {
  fastify: '1.x',
  decorators: {
    fastify: ['plugin1', 'plugin2'],
    reply: ['compress']
  },
  dependencies: ['plugin1-name', 'plugin2-name']
})
```

### Output
```
--> generated package.json in D:\Repositories\fastify-cli\myPlugin
--> reading package.json in D:\Repositories\fastify-cli\myPlugin
generated README.md
--> README for plugin myPlugin generated successfully
```
<pre>
# myPlugin

Short description of the purpose of your plugin.

## Install

```
npm install myPlugin --save
yarn add myPlugin
```

## Example

```js
const fastify = require('fastify')()
fastify.register(require('myPlugin'))
fastify.listen(3000)
```

You can also start any Fastify plugin with the [Fastify-cli](https://github.com/fastify/fastify-cli):

```
fastify start plugin.js
```

### Accessibility

- [X] Accessible in the same context where you require them
- [ ] Accessible only in a child context

#### Fastify

- plugin1
- plugin2

#### Reply

- compress

## Dependencies

- plugin1-name
- plugin2-name

</pre>


cli.js
```diff
@@ -10,9 +10,11 @@ const help = require('help-me')({
 })
 const start = require('./start')
 const generate = require('./generate')
+const generatePlugin = require('./generate-plugin')
 
 commist.register('start', start.cli)
 commist.register('generate', generate.cli)
+commist.register('generate-plugin', generatePlugin.cli)
```
mcollina: Can you please check that commist [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) keeps picking up `fastify gen` to generate apps, and maybe `fastify plugin` to generate a plugin?

Maybe we could even change the commands and have `fastify generate-application` or `fastify app` to generate and application, and `fastify plugin` to generate a plugin.

generate-plugin.js
```diff
@@ -0,0 +1,155 @@
...
+
+const colors = [l => l, chalk.green, chalk.red]
+
+function cli (args) {
+  const opts = argv(args)
+
+  const dir = process.cwd()
+
+  if (opts._.length !== 1) {
+    log('error', 'Error: Missing the required file parameter\n')
```
mcollina: Should we omit `Error:`?

cli.js
StarpTech: I'm a bit confused. My goal is to generate a README from an existing plugin. I don't want to implement a plugin generator.

mcollina: ok then, it should be something like "gen-readme"

StarpTech: `generate-app` will still work because of Levenshtein distance

generate.js
```diff
@@ -78,21 +74,11 @@ function generate (dir, log, cb) {
...
-const colors = [
-  (l) => l,
-  chalk.green,
-  chalk.red
-]
-
 function cli (args) {
   const opts = argv(args)
 
+  log('info', 'Execute command: gen-app')
```
StarpTech: I added it to each command due to Levenshtein disance. Its not always clear which command is executed when we have a common prefix like `gen`

mcollina:
Maybe we can just call one `generate` and the other one `readme`? would that make it simpler? The README would need to be adjusted as well.

StarpTech:
@delvedor why it is useless when we are trying to provide a concise description of the plugin?

delvedor:
Because usually, I create everything at the same time.
Having a tool that generates all the minimum files needed to have a working plugin is more valuable than creating manually everything and then generate the README, because you are gonna need the other files either way.

StarpTech:
See it as a first step. I think it can already be useful especially for newcomers.

* Implement your plugin
* Generate readme
* Add tests
= 🥇

delvedor:
Again, I think is more useful (especially for newcomers) to have all the files at the same time, as a full example (entry point, test, **and** readme).
In the same way we are doing for the project generator.

StarpTech:
I got your point but I never talked about a plugin generator.

> It should encourage plugin developers to provide a concise plugin readme of a fastify plugin. Consumers can immediately check what decorators a plugin provides, on which dependencies it relies and more ... All the metainformation can be extracted from a fastify plugin

Nothing more or less. Feel free to create PR to create a plugin generator 😄

PR #158, opened 2019-02-25
Support for Fastify v2

mcollina:


mcollina:
That's built in: https://github.com/fastify/fastify-cli/blob/5719a87732e14c28f741b10180744329ca45ba2f/generate.js#L49.

We just need to bump the version.

cemremengu:
@mcollina will we bump Fastify in a separate PR ?

PR #182, opened 2019-06-22
chore: warns the user appropriately for non-existent files

### LINKED ISSUE #181: Throws error on providing a file that doesn't exist with start command
`fastify start <file>` throws an error if the respective file doesn't exist:-

Screenshot from 2019-06-22 23-02-21

It seems that this was intentional as from:-
 https://github.com/fastify/fastify-cli/blob/f8ca00c1066c779b822b315a07162098fe41b161/start.js#L91

IMO it is good to warn the user appropriately :thinking:

jamesgeorge007:
Fixes #181

Shows up a warning rather than throwing the error.

### DIFF
--- a/start.js
+++ b/start.js
@@ -92,7 +92,7 @@ function runFastify (args, cb) {
   const filePath = path.resolve(process.cwd(), opts._[0])
 
   if (!fs.existsSync(filePath)) {
-    return module.exports.stop('', `${opts._[0]} doesn't exist within ${process.cwd()}`)
+    return module.exports.stop(null, `${opts._[0]} doesn't exist within ${process.cwd()}`)
   }
 
   let file = null
--- a/start.js
+++ b/start.js
@@ -87,8 +87,7 @@ function runFastify (args, cb) {
   loadModules(opts)
 
   if (!fs.existsSync(path.resolve(process.cwd(), opts._[0]))) {
-    console.log(`${opts._[0]} doesn't exist within ${process.cwd()}`)
-    return
+    return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
   }
 
   let file = require(path.resolve(process.cwd(), opts._[0]))
--- a/start.js
+++ b/start.js
@@ -78,6 +78,7 @@ function stop (error, warn) {
   }
   if (warn) {
     console.log(warn)
+    process.exit(1)
   }
   process.exit()
 }
--- a/start.js
+++ b/start.js
@@ -86,11 +86,13 @@ function runFastify (args, cb) {
 
   loadModules(opts)
 
-  if (!fs.existsSync(path.resolve(process.cwd(), opts._[0]))) {
-    return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
+  const filePath = path.resolve(process.cwd(), opts._[0])
+
+  if (!fs.existsSync(filePath)) {
+    return module.exports.stop(new Error(`${opts._[0]} doesn't exist within ${process.cwd()}`))
   }
 
-  let file = require(path.resolve(process.cwd(), opts._[0]))
+  const file = require(filePath)
 
   if (file.length !== 3 && file.constructor.name === 'Function') {
     return module.exports.stop(new Error('Plugin function should contain 3 arguments. Refer to ' +
--- a/start.js
+++ b/start.js
@@ -77,7 +77,7 @@ function stop (error, warn) {
     process.exit(1)
   }
   if (warn) {
-   console.log(warn)
+    console.log(warn)
   }
   process.exit()
 }
@@ -95,7 +95,13 @@ function runFastify (args, cb) {
     return module.exports.stop('', `${opts._[0]} doesn't exist within ${process.cwd()}`)
   }
 
-  const file = require(filePath)
+  let file = null
+
+  try {
+    file = require(filePath)
+  } catch (e) {
+    return module.exports.stop(e)
+  }
 
   if (file.length !== 3 && file.constructor.name === 'Function') {
     return module.exports.stop(new Error('Plugin function should contain 3 arguments. Refer to ' +
--- a/start.js
+++ b/start.js
@@ -86,12 +86,12 @@ function runFastify (args, cb) {
 
   loadModules(opts)
 
-  var file = null
-  try {
-    file = require(path.resolve(process.cwd(), opts._[0]))
-  } catch (e) {
-    return module.exports.stop(e)
+  if (!fs.existsSync(path.resolve(process.cwd(), opts._[0]))) {
+    console.log(`${opts._[0]} doesn't exist within ${process.cwd()}`)
+    return
   }
+
+  let file = require(path.resolve(process.cwd(), opts._[0]))
 
   if (file.length !== 3 && file.constructor.name === 'Function') {
     return module.exports.stop(new Error('Plugin function should contain 3 arguments. Refer to ' +
--- a/start.js
+++ b/start.js
@@ -71,10 +71,13 @@ function start (args, cb) {
   return runFastify(args, cb)
 }
 
-function stop (error) {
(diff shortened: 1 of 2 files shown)


jamesgeorge007:
@mcollina How should the tests be refactored so as to cope up with the respective update :thinking:

start.js
```diff
@@ -86,13 +86,13 @@ function runFastify (args, cb) {
 
   loadModules(opts)
 
-  var file = null
-  try {
-    file = require(path.resolve(process.cwd(), opts._[0]))
-  } catch (e) {
-    return module.exports.stop(e)
+  if (!fs.existsSync(path.resolve(process.cwd(), opts._[0]))) {
+    console.log(`${opts._[0]} doesn't exist within ${process.cwd()}`)
```
xtx1130:
```suggestion
    return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
```

jamesgeorge007:
@mcollina Can you throw light on writing the unit test for the respective case?

mcollina:
You can mock `module.exports.stop` and verify it's getting called.

jamesgeorge007:
@mcollina It seems that 4 test cases are failing. 

How should be the following error message be refactored :thinking: 

https://github.com/fastify/fastify-cli/blob/f8ca00c1066c779b822b315a07162098fe41b161/test/start.test.js#L151

mcollina:
You need to change https://github.com/fastify/fastify-cli/blob/f8ca00c1066c779b822b315a07162098fe41b161/start.js#L74 so that it accept a second parameter (a string), and you print that string in `stop` if the error is not provided. Then, you need to change your PR so that the message is passed as a second argument instead.

start.js
```diff
@@ -86,13 +86,14 @@ function runFastify (args, cb) {
...
 
-  var file = null
-  try {
-    file = require(path.resolve(process.cwd(), opts._[0]))
-  } catch (e) {
-    return module.exports.stop(e)
+  const filePath = path.resolve(process.cwd(), opts._[0])
+
+  if (!fs.existsSync(filePath)) {
+    return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
```
xtx1130:
```suggestion
    return module.exports.stop(new Error(`${opts._[0]} doesn't exist within ${process.cwd()}`))
```
I’m sorry for my previous suggestion. There must be an instance of Error, So that the test case can capture `error.message`

jamesgeorge007:
@xtx1130 Can you give me a better picture on fixing those failing unit tests :thinking:

jamesgeorge007:
@xtx1130 Can you draw a better picture regarding the failing tests and how should it be fixed up :thinking:

xtx1130:
@jamesgeorge007 I have got wrong, you need to accept mcollina's suggest, add another param to ensure stop support warning, such as :
```js
function stop (error, warn) {
  if (error) {
    console.log(error)
    process.exit(1)
  }
  if (warn) {
   console.log(warn)
  }
  process.exit()
}
```
And you need to change:
```js
return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
```
to 
```js
return module.exports.stop('', `${opts._[0]} doesn't exist within ${process.cwd()}`)
```
The test case should change to:
```js
test('should throw on file not found', t => {
  t.plan(1)

  const oldStop = start.stop
  t.tearDown(() => { start.stop = oldStop })
  start.stop = function (err, warn) {
   // test case changes
    t.ok(/.*not-found doesn't exist within/.test(warn), warn)
  }

  const argv = [ '-p', getPort(), './data/not-found.js' ]
  start.start(argv)
})
```

xtx1130:
@jamesgeorge007 There still have some test cases you need to deal with. You can run `npm run test` in your local environment to make sure the test cases is correct.

jamesgeorge007:
@xtx1130 I'm not able to figure out what's going wrong with the currently failing tests. Any sort of help is appreciated :clap:

start.js
```diff
@@ -86,13 +89,14 @@ function runFastify (args, cb) {
...
-    file = require(path.resolve(process.cwd(), opts._[0]))
-  } catch (e) {
-    return module.exports.stop(e)
+  const filePath = path.resolve(process.cwd(), opts._[0])
+
+  if (!fs.existsSync(filePath)) {
+    return module.exports.stop('', `${opts._[0]} doesn't exist within ${process.cwd()}`)
   }
 
+  const file = require(filePath)
```
xtx1130:
```suggestion
  let file = null
  try {
    file = require(filePath)
  } catch (e) {
    return module.exports.stop(e)
  }
```

xtx1130: You just ensure the `filePath` exists in folder, how about the require files in `filePath`? So you need to add this try catch to ensure `require(filePath)` run correctly.

jamesgeorge007:
cc @xtx1130

test/start.test.js
```diff
@@ -160,8 +161,8 @@ test('should throw on package not found', t => {
 
   const oldStop = start.stop
   t.tearDown(() => { start.stop = oldStop })
-  start.stop = function (err) {
-    t.ok(/Cannot find module.*unknown-package/.test(err.message), err.message)
+  start.stop = function (err, warn) { // eslint-disable-line
```
Eomm: instead of disable the linting you could test that the warn parameter is empty or not

jamesgeorge007:
cc @Eomm

start.js
```diff
@@ -86,9 +89,16 @@ function runFastify (args, cb) {
 
   loadModules(opts)
 
-  var file = null
+  const filePath = path.resolve(process.cwd(), opts._[0])
+
+  if (!fs.existsSync(filePath)) {
+    return module.exports.stop('', `${opts._[0]} doesn't exist within ${process.cwd()}`)
```
Eomm:
```suggestion
    return module.exports.stop(null, `${opts._[0]} doesn't exist within ${process.cwd()}`)
```

jamesgeorge007:
cc @Eomm

jamesgeorge007:
cc @mcollina

test/start.test.js
```diff
@@ -147,8 +147,9 @@ test('should throw on file not found', t => {
 
   const oldStop = start.stop
   t.tearDown(() => { start.stop = oldStop })
-  start.stop = function (err) {
-    t.ok(/Cannot find module.*not-found/.test(err.message), err.message)
+  start.stop = function (err, warn) { // eslint-disable-line
```
mcollina: can you add an assertion that `err` is `nulll`

test/start.test.js
```diff
@@ -160,8 +161,8 @@ test('should throw on package not found', t => {
 
   const oldStop = start.stop
   t.tearDown(() => { start.stop = oldStop })
-  start.stop = function (err) {
-    t.ok(/Cannot find module.*unknown-package/.test(err.message), err.message)
+  start.stop = function (err, warn) {
+    t.ok(/Cannot find module 'unknown-package'/.test(err.message), err.message)
```
mcollina: can you add an assertion on `warn` being `null` or `undefined`?

test/start.test.js
```diff
@@ -173,7 +174,7 @@ test('should throw on parsing error', t => {
 
   const oldStop = start.stop
   t.tearDown(() => { start.stop = oldStop })
-  start.stop = function (err) {
+  start.stop = function (err, warn) { // eslint-disable-line
```
mcollina: can you add an assertion on `warn` being `null` or `undefined`?

start.js
```diff
@@ -71,11 +71,14 @@ function start (args, cb) {
...
 }
 
-function stop (error) {
+function stop (error, warn) {
   if (error) {
     console.log(error)
     process.exit(1)
   }
+  if (warn) {
+    console.log(warn)
```
mcollina: I think we should exit with code 1 even in case of a `warn`.

jamesgeorge007:
cc @mcollina

PR #193, opened 2019-08-10
Add command to print routes

francisbrito:
Hello,

I found myself needing a way to see a list (or tree) breakdown of all the routes my project had and thought it might be useful to have a command in `fastify-cli` that does just that.

This is essentially a CLI wrapper for `fastify.printRoutes`.

It behaves similarly to `fastify start` command except it relies on `fastify.ready()` rather than `fastify.listen()`, as such, there are a few function that are identical in both commands and could probably be abstracted and reused. If it seems reasonable, I could do it in the same PR. 

The reason I didn't write these abstractions is I preferred not to introduce too many changes to the project structure in order not to make this PR too invasive.

### DIFF
--- a/print-routes.js
+++ b/print-routes.js
@@ -1,0 +1,116 @@
+#! /usr/bin/env node
+
+'use strict'
+
+require('dotenv').config()
+
+const assert = require('assert')
+const path = require('path')
+const fs = require('fs')
+const resolveFrom = require('resolve-from')
+const parseArgs = require('./args')
+const log = require('./log')
+
+let Fastify = null
+
+function loadModules (opts) {
+  try {
+    const basedir = path.resolve(process.cwd(), opts._[0])
+
+    Fastify = require(resolveFrom.silent(basedir, 'fastify') || 'fastify')
+  } catch (e) {
+    module.exports.stop(e)
+  }
+}
+
+function showHelp () {
+  console.log(fs.readFileSync(path.join(__dirname, 'help', 'print-routes.txt'), 'utf8'))
+  return module.exports.stop()
+}
+
+function printRoutes (args, cb) {
+  const opts = parseArgs(args)
+  if (opts.help) {
+    return showHelp()
+  }
+
+  if (opts._.length !== 1) {
+    console.error('Missing the required file parameter\n')
+    return showHelp()
+  }
+
+  // we start crashing on unhandledRejection
+  require('make-promises-safe')
+
+  loadModules(opts)
+
+  return runFastify(opts, cb)
+}
+
+function runFastify (opts, cb) {
+  const filePath = path.resolve(process.cwd(), opts._[0])
+  cb = cb || assert.ifError
+
+  if (!fs.existsSync(filePath)) {
+    return module.exports.stop(`${opts._[0]} doesn't exist within ${process.cwd()}`)
+  }
+
+  let file = null
+
+  try {
+    file = require(filePath)
+  } catch (e) {
+    return module.exports.stop(e)
+  }
+
+  if (file.length !== 3 && file.constructor.name === 'Function') {
+    return module.exports.stop(new Error('Plugin function should contain 3 arguments. Refer to ' +
+      'documentation for more information.'))
+  }
+  if (file.length !== 2 && file.constructor.name === 'AsyncFunction') {
+    return module.exports.stop(new Error('Async/Await plugin function should contain 2 arguments.' +
+      'Refer to documentation for more information.'))
+  }
+
+  const fastify = Fastify(opts.options)
+
+  const pluginOptions = {}
+  if (opts.prefix) {
+    pluginOptions.prefix = opts.prefix
+  }
+
+  fastify
+    .register(file, pluginOptions)
+    .ready((err) => {
+      if (err) {
+        return cb(err, fastify)
+      }
+
+      log('debug', fastify.printRoutes())
+
+      cb(null, fastify)
+    })
+
+  return fastify
+}
+
+function stop (message) {
+  if (message instanceof Error) {
+    console.log(message)
+    process.exit(1)
+  } else if (message) {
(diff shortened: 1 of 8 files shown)


print-routes.js
```diff
@@ -0,0 +1,116 @@
...
+
+function cli (args) {
+  printRoutes(args)
+}
+
+module.exports = { cli, stop, printRoutes }
+
+if (require.main === module) {
+  cli(process.argv.slice(2))
+}
```
mcollina: Can you refactor the logic in this file and the start command to reduce duplication?

print-routes.js
```diff
@@ -0,0 +1,116 @@
...
+    process.exit(1)
+  } else if (message) {
+    console.log(`Warn: ${message}`)
+    process.exit(1)
+  }
+  process.exit()
+}
+
+function cli (args) {
+  printRoutes(args)
```
mcollina: I think we should call fastify.close() after having printed the routes. It’s important so the application would stop gracefully.

print-routes.js
francisbrito: Sure thing!

I have created `util.js` just for this. Please have a look and let me know what you think.

print-routes.js
francisbrito: Good catch.

I added a call to `fastify.close` in my latest commit.

francisbrito:
In addition to applying the improvements highlighted in the review, I added the an entry for the `print-routes` command to the `help` command output.

Also fixed tests failing in Node 6. (I hadn't realized Travis CI check was failing).

start.js
```diff
@@ -4,34 +4,33 @@
...
-    Fastify = require(resolveFrom.silent(basedir, 'fastify') || 'fastify')
-    fastifyPackageJSON = require(resolveFrom.silent(basedir, 'fastify/package.json') || 'fastify/package.json')
+    Fastify = fastifyModule
+    fastifyPackageJSON = fastifyPkg
   } catch (e) {
     module.exports.stop(e)
   }
 }
 
 function showHelp () {
```
xtx1130: I think it is unnecessary to make `showHelpForCommand` inside `showHelp`. You can use `exit()` inside `showHelpForCommand`, and delete `showHelp` there.

util.js
```diff
@@ -0,0 +1,72 @@
...
+
+function requireFastifyForModule (modulePath) {
+  try {
+    const basedir = path.resolve(process.cwd(), modulePath)
+    const module = require(resolveFrom.silent(basedir, 'fastify') || 'fastify')
+    const pkg = require(resolveFrom.silent(basedir, 'fastify/package.json') || 'fastify/package.json')
+
+    return { module, pkg }
+  } catch (e) {
+    console.log('Warn: unable to load fastify module')
```
xtx1130:
```suggestion
    exit('unable to load fastify module')
```

util.js
```diff
@@ -23,8 +23,8 @@ function requireFastifyForModule (modulePath) {
 
     return { module, pkg }
   } catch (e) {
-    console.log('Warn: unable to load fastify module')
-    throw e
+    console.log('unable to load fastify module')
+    exit()
```
xtx1130: I think you can use `exit(message)` instead of `console.log(message)`

util.js
```diff
@@ -63,8 +63,9 @@ function showHelpForCommand (commandName) {
 
   try {
     console.log(fs.readFileSync(helpFilePath, 'utf8'))
+    exit()
```
xtx1130: And there

francisbrito: I think changing this line to be `exit(fs.readFileSync(helpFilePath, 'utf8'))` would cause help command to return a non-zero exit code. Is this OK?

I might be misinterpreting your change request, so please do let me know if so is the case.

xtx1130: @francisbrito I am sorry for my mistake, I wanna add comment on [Line67](https://github.com/fastify/fastify-cli/pull/193/files#diff-63f3aa32c5cc5d42a250b081693e9dc2R67). I think is unnecessary to tell users two messages about the error

util.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function showHelpForCommand (commandName) {
+  const helpFilePath = path.join(__dirname, 'help', `${commandName}.txt`)
+
+  try {
+    console.log(fs.readFileSync(helpFilePath, 'utf8'))
+    exit()
+  } catch (e) {
+    console.log(`unable to get help for command "${commandName}"`)
```
xtx1130: Here🙋‍♂️。
```suggestion
    exit(new Error(`unable to get help for command "${commandName}"`))
```

francisbrito: My reasoning for adding the extra `console.log` comment was adding some context on what is the underlying cause of the error. But thinking it better, if a command errors when the user passes the `--help` flag the error would be out of the hands of their hands.

print-routes.js
```diff
@@ -0,0 +1,86 @@
...
+
+let Fastify = null
+
+function loadModules (opts) {
+  try {
+    Fastify = requireFastifyForModule(opts._[0]).module
+  } catch (e) {
+    module.exports.stop(e)
+  }
+}
```
mcollina: I think this can be abstracted away and shared with start.

francisbrito: Like this?

```js
// util.js
function tryRequireFastifyForModuleOrExit(modulePath) {
  try {
    return requireFastifyForModule(modulePath);
  } catch (e) {
    exit(e);
  }
}
```

mcollina: I was wrong. It's fine as it is.

test/start.test.js
```diff
@@ -209,17 +209,21 @@ test('should start the server with an async/await plugin', t => {
   })
 })
 
-test('should exit without error on help', t => {
-  t.plan(1)
+t.only('should exit without error on help', t => {
```
Eomm:
```suggestion
test('should exit without error on help', t => {
```

Issue #200, opened 2019-08-27
Can not run fastify cli in watch mode with node inspector


pyxarez:
## 🐛 Bug Report

Cli is not working in --watch mode with node inspect on the board
It throws the following log: `Starting inspector on 127.0.0.1:9230 failed: address already in use`

## To Reproduce

Paste your code here:

```js
node --inspect-brk=9230 ./node_modules/.bin/fastify start --watch -l info -P app.js
```

## Expected behavior

It should work in --watch mode with node inspect on the board

## Your Environment

- *node version*: 10
- *fastify version*: >=2.0.0
- *os*: Mac

mcollina:
This happens because we are forking the project: https://github.com/fastify/fastify-cli/blob/master/lib/watch/fork.js. We might want to do something about this, as it's a rather convenient feature.

Can you do the same with Nodemon for example? How do they implement/support it?

pyxarez:
> Can you do the same with Nodemon for example? How do they implement/support it?
I don't fully understand the question
Yea, I can try to do the same with nodemon, but it requires [migrating](https://github.com/fastify/fastify-cli) from cli, making my own server.js and leads to loss of fastify-cli functionality like logging, default setup, etc.
It will look like this `nodemon --inspect=9230 server.js`

mcollina:
You answered the question! `nodemon` implements it via a `--inspect` flag. Would you like to send a PR that adds that to fastify-cli as well?

pyxarez:
Hmm, that's an interesting idea
Yeah, I can try to do that, but I'm new in fastify and I think it'll take some time
If you do not need super fast fix that's the task for me

mcollina:
Go ahead!

xtx1130:
I am sure `child_process.fork()` caused this, you need add options `{execArgv: []}` to `child_process.fork()`

pyxarez:
@xtx1130 , I tried you suggestion and it's actually work!
But will this not lead to errors in the case when we really want to use execArgv?

pyxarez:
I admit that that's an internal usage of `fork` and there is no case when someone would like to parametrise that part from outisde

HighSoftWare96:
I'm going to make a PR of this if you want... I was able to manage that with `execArgv` parameter and adding a `debug` flag to the CLI.
![image](https://user-images.githubusercontent.com/22793967/63942214-8a2d9580-ca6d-11e9-9f39-b159893d8af9.png)
[...]

PR #202, opened 2019-08-29
Debug mode for Fastify CLI Start

HighSoftWare96:
#### Feature
With this PR you should be able to run a FastifyCLI-generated app with NodeJS inspector on a chosen port. The added flags are: `-d` (`--debug`) for enabling the debug mode and `-I` (`--debug-port`) for choosing a debug port. Works even in `--watch` mode.

#### What was done
- Programmatically call `inspector.open` right before running Fastify for the current process when debug mode active. Refs: https://nodejs.org/api/inspector.html
- Updated readme.md documentation

#### Contributions
This PR was created referring this issue: https://github.com/fastify/fastify-cli/issues/200. Thanks also to @pyxarez and @xtx1130. Feel free to contribute.

#### Checklist

- [-] run `npm run test` and `npm run benchmark`
- [-] tests and/or benchmarks are included
- [-] documentation is changed or added
- [-] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -15,7 +15,7 @@ const forkPath = path.join(__dirname, './fork.js')
 
 const emitter = new EventEmitter()
 
-let allStop = false
+const allStop = false
 let childs = []
 
 const stop = (watcher = null, err = null) => {
@@ -26,12 +26,11 @@ const stop = (watcher = null, err = null) => {
   childs = []
   if (err) { console.log(chalk.red(err)) }
   if (watcher) {
-    allStop = true
     watcher.close()
   }
 }
 
-const watch = function (args, ignoreWatch) {
+const watch = function (args, ignoreWatch, debugMode, debugPort) {
   process.on('uncaughtException', () => {
     stop()
     childs.push(run('restart'))
@@ -40,11 +39,11 @@ const watch = function (args, ignoreWatch) {
   const run = (event) => {
     const childEvent = { childEvent: event }
     const env = Object.assign({}, process.env, childEvent)
-    console.log("Forking");
-    const execArgv = [];
 
-    if (args.debug) {
-      execArgv.push('--inspect-brk=' + (args.debugPort || 9229))
+    const execArgv = []
+
+    if (debugMode) {
+      execArgv.push(`--inspect=${debugPort || 9229}`)
     }
 
     const _child = cp.fork(forkPath, args, {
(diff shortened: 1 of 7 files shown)


start.js
```diff
@@ -13,6 +13,7 @@ const listenAddressDocker = '0.0.0.0'
 const watch = require('./lib/watch')
 const parseArgs = require('./args')
 const { exit, requireFastifyForModule, requireServerPluginFromPath, showHelpForCommand } = require('./util')
+const inspector = require('inspector')
```
mcollina: We are still supporting node 6, and this is not available there. Would you mind loading this module only in that case?

HighSoftWare96: Sure, I undestand it is a experimental feature of nodejs but it was the simplest way to do that :)

mcollina: That's brilliant, definitely +1!

HighSoftWare96:
> uld you mind adding a quick example that adds a test using

I'm not confident with this tool. What should I do?

mcollina:
I would do something like:

```js
proxyquire('../start', {
  inspector: {
    open () {
      t.pass('open called')
    }
  }
})
```

lib/watch/index.js
```diff
@@ -26,7 +26,6 @@ const stop = (watcher = null, err = null) => {
   childs = []
   if (err) { console.log(chalk.red(err)) }
   if (watcher) {
-    allStop = true
     watcher.close()
```
xtx1130: Why you delete `allStop = true` here?

HighSoftWare96: My bad, I don't know where it comes from

start.js
```diff
@@ -101,6 +101,14 @@ function runFastify (args, cb) {
     pump(pinoColada, process.stdout, assert.ifError)
   }
 
+  if (opts.debug) {
+    if (process.version.match(/v[0-6]\..*/g)) {
+      console.warn('Fastify debug mode not compatible with Node.js version < 6')
```
xtx1130:
```suggestion
      stop('Fastify debug mode not compatible with Node.js version < 6')
```

HighSoftWare96:
@xtx1130 would you mind complete the PR with the tests referred by @mcollina? I have not enough time and so the PR will be completed

start.js
HighSoftWare96: Should be fine, thanks!

xtx1130:
@HighSoftWare96 It soon be a weekend in china. I can finish it after 2 days if you don't mind.
@mcollina Would you mind merge this pr into master, and I will add a test-case for this in another pr?

PR #215, opened 2020-01-09
Add -watch option to dev task

olafkrueger:
Wouldn't it makes sense to add the `-watch` option to the dev task?


olafkrueger:
> Would you mind to fix it?

What a great first pull request of mine ;-)
Sorry for breaking things, seems to be fixed...

Issue #214, opened 2020-01-06
Generated project does not show error stacktraces even in DEV mode


vitalets:
`fastify generate` creates npm script `"dev": "fastify start -l info -P app.js"` to run app in development mode. But in that mode no stacktraces are shown in logs: 

![image](https://user-images.githubusercontent.com/1473072/71835951-f24b5280-30c3-11ea-850a-a7e1b7ccbf7b.png)

This obviously unsuitable for development. 
I've researched `pino-colada` logger and found that it [should show stacktraces](https://github.com/lrlna/pino-colada/blob/772b000deb0fff3c918ac974543a3b89f0dd9340/index.js#L112) only in `error` or `fatal` mode. But changing run script to `fastify start -l error -P app.js` also does not help. It's because `pino-colada` expects stacktrace in `obj.stack` but fastify writes it to `obj.err.stack`.

Related discussion: https://github.com/lrlna/pino-colada/pull/30

mcollina:
I think we should switch to https://github.com/pinojs/pino-pretty. @jsumners @davidmarkclements @delvedor what do you think?

jsumners:
Sounds reasonable to me. I believe Lrlna has moved on from `pino-colada`.

mcollina:
Would you like to open a PR for that @vitalets?

vitalets:
I've checked pino-pretty and frankly speaking it's default output seems not very pretty for me:
1. very verbose
2. error stack is very nested and not colored in red

Compare this screenshot with pino-colada output in my first message:
![image](https://user-images.githubusercontent.com/1473072/71882205-0f743580-3145-11ea-9f63-17d95926aea1.png)

For me personally pino-colada looks more pretty ) Maybe just fix on their side to check `obj.err.stack` along with `obj.stack`? @lrlna what do you think?

I you anyway going to switch to pino-pretty - I think we should negotiate the format. Because all current users of fastify-cli will be surprised with such verbose format after concise pino-colada.

lrlna:
@vitalets if you want to make a PR to `pino-colada` to also check for `obj.err.stack`, I'd happily accept it. There is only [one place](https://github.com/lrlna/pino-colada/blob/master/index.js#L112-L115) where the check happens.

vitalets:
[...]

PR #216, opened 2020-01-14
Update pino-colada to 1.5.1, see #214

vitalets:
See #214


mcollina:
This is not needed, the new version is already covered in the version range. Update your dependencies.

vitalets:
ok 👍

PR #222, opened 2020-02-11
add command to render interactive docs to the terminal

GlenTiki:
This PR adds the ability for you to view the documentation for fastify in your terminal. By default, `fastify-cli` attempts to render the documentation for the fastify version installed in the current working directory node_modules folder, however, if none are found, it should fall back to rendering the documentation for the version that fastify-cli depends on.

The documentation is rendered using an interactive terminal session that you can navigate with your arrow keys.

run `fastify docs` to get started.

#### Checklist

Screenshots:

Screenshot 2020-02-11 at 17 55 40
Screenshot 2020-02-11 at 17 56 40
Screenshot 2020-02-11 at 17 56 50

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -12,10 +12,12 @@ const start = require('./start')
 const generate = require('./generate')
 const generateReadme = require('./generate-readme')
 const printRoutes = require('./print-routes')
+const renderDocs = require('./docs')
 
 commist.register('start', start.cli)
 commist.register('generate', generate.cli)
 commist.register('readme', generateReadme.cli)
+commist.register('docs', renderDocs)
 commist.register('help', help.toStdout)
 commist.register('version', function () {
   console.log(require('./package.json').version)
--- a/docs.js
+++ b/docs.js
@@ -1,0 +1,127 @@
+const blessed = require('blessed')
+const contrib = require('blessed-contrib')
+const walkup = require('walk-up')
+const fs = require('fs').promises
+const { join } = require('path')
+
+function findDocs () {
+  return new Promise((resolve, reject) => {
+    const searchDir = 'node_modules/fastify/docs'
+    walkup(process.cwd(), searchDir, (err, results) => {
+      if (err) {
+        return reject(err)
+      }
+
+      if (results.found) {
+        return resolve(join(results.path, searchDir))
+      } else {
+        walkup(__dirname, searchDir, (err, results) => {
+          if (err) {
+            return reject(err)
+          }
+
+          if (results.found) {
+            return resolve(join(results.path, searchDir))
+          } else {
+            return resolve()
+          }
+        })
+      }
+    })
+  })
+}
+
+async function renderDocs () {
+  const docsBase = await findDocs()
+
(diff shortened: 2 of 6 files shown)


docs.js
```diff
@@ -0,0 +1,127 @@
+const blessed = require('blessed')
```
mcollina: please add 'use strict'

cli.js
```diff
@@ -12,10 +12,12 @@ const start = require('./start')
 const generate = require('./generate')
 const generateReadme = require('./generate-readme')
 const printRoutes = require('./print-routes')
+const renderDocs = require('./docs')
 
 commist.register('start', start.cli)
 commist.register('generate', generate.cli)
 commist.register('readme', generateReadme.cli)
+commist.register('docs', renderDocs)
```
mcollina: We'll need to disable this in Node < 10 I think

docs.js
```diff
@@ -0,0 +1,127 @@
...
+const contrib = require('blessed-contrib')
+const walkup = require('walk-up')
+const fs = require('fs').promises
+const { join } = require('path')
+
+function findDocs () {
+  return new Promise((resolve, reject) => {
+    const searchDir = 'node_modules/fastify/docs'
```
mcollina: You might want to use `require.resolve` instead. Yarn pnp and pnpm would not find this otherwise.

GlenTiki: Can require.resolve be used for resolving a directory? I thought it could only resolve .js files?

cli.js
GlenTiki: Fair, I'll add a check in the render docs fn

mcollina: you cannot require that file on node 6 and possibly 8. We might just bump the major anyway, would you be ok with that? (This might land on a next branch then).

docs.js
mcollina: you can resolve the entry point and then move in that folded

GlenTiki: Fair, I'll do this! Ty

cli.js
GlenTiki: I'm happy with that 😊

GlenTiki: I've got around this by removing async/await, and just using the sync fs funcs, sync should be fine for this use-case I think

GlenTiki:
@mcollina I was thinking the same. I'll try recompose the code to see if there is an easy way to achieve it, but I think it will be a pain to test.

GlenTiki:
Yeah - after trying to compose this for testing, it's quite awkward and not much in the way of the user code is being tested. If you're happy to merge feel free to do so.

GlenTiki:
@Eomm The primary reason I created this was for offline usage 😄 When I have internet I would likely be on the website anyway haha

lmammino:
Awesome feature 😱

lmammino:
I wonder whether, in addition to terminal docs, it could be nice to open the docs on the browser. I still have to fully flesh this thought out, but there might be ways to generalize some of the code used to build the docs in the website in a shared library that we can import here as well. Thoughts @GlenTiki?

Came to think about this after running `rustup docs` 🤔

Issue #131, opened 2018-11-19
Support for TypeScript template and eject command 


fox1t:
Hi, guys. It will be nice to include in fastify-cli support for TypeScript projects. 
In order to generate a TS  project we need to change some of the current fastify-cli's commands and to add one new:
 * `start`       start a server -> here we can add support for .ts files  (auto-discover file extension), add default tsconfig.json file with default compilation rules and use ts-node with inline sourcemap generation to run the code. Notes: Running .ts is not intended for production environment, .ts files will be inside `/src` folder, as this is a TS standard defacto.
  * `generate`   generate a new project -> we need to add new arg `--typescript` (`-ts`) to generate TypeScript project with all dependencies and to add to `package.json` typings and typescript compiler. In addition we need (just one?) new npm-script to build .ts code.
  * `eject`   ejects the app (__new command__) ->  creates server.ts file and compiles all TypeScript code from src to `build` (or `dist`) folder and modifies scripts to use newly generated paths. Note: this command could be used also by JS projects.

In addition to that, what do you think to add one more new npm-script for running debugger `in package.json`? It would be nice to have it ready inside package.json.

I am looking forward for ideas and I am ready to implement this feature,

delvedor:
> generate generate a new project -> we need to add new arg --typescript (-ts) to generate TypeScript project with all dependencies and to add to package.json typings and typescript compiler. In addition we need (just one?) new npm-script to build .ts code.

I'm not sure you can do `-ts`, as it should be read as `-t -s` by the parser.

> eject ejects the app (new command) -> creates server.ts file and compiles all TypeScript code from src to build (or dist) folder and modifies scripts to use newly generated paths. Note: this command could be used also by JS projects.

This means that you should "eject" the app everytime you need to deploy to production? If so I don't like it too much, it feels more complex than it should be.

`fastify-cli` is intended to work both on prod and development, and it should continue to do so in my opinion.

> In addition to that, what do you think to add one more new npm-script for running debugger in package.json? It would be nice to have it ready inside package.json.

Can you post an example?

cemremengu:
Sounds very nice actually! Currently we do eject manually (see README) so that would be a nice feature to have as well.

I would be happy to review a PR as long as it doesnt break current functionality for javascript.

mcollina:
I'm +1 on adding support for ts.

cemremengu:
My comment is gone I think I deleted it mistakenly on the phone but I am +1 as well.

[...]

PR #227, opened 2020-02-28
(#131) Created typescript template and added generate command for typ…

alemagio:
…escript option (#131)

#### Checklist

- [x] run `npm run test`

It is a draft, I followed the official [TypeScript-Node-Starter](https://github.com/Microsoft/TypeScript-Node-Starter).
If you like the base idea I will refactor it a bit avoiding some code duplications and add tests and doc.


mcollina:
Good work! Keep going, I’ve left a few comments!

alemagio:
> Good work! Keep going, I’ve left a few comments!

Hi, i'm sorry i can't find your comments.

generate.js
```diff
@@ -74,8 +74,74 @@ function generate (dir, cb) {
...
+        return cb(err)
+      }
+
+      pkg.main = 'app.ts'
+
+      pkg.scripts = Object.assign(pkg.scripts || {}, {
+        test: 'tap test/**/*.test.js',
+        start: 'npm run build && fastify start -l info dist/app.js',
+        'build:ts': 'tsc',
+        dev: 'tsc && concurrently -k -p "[{name}]" -n "TypeScript,App" -c "yellow.bold,cyan.bold"  "tsc -w" "../cli.js start -w -l info -P dist/app.js"'
```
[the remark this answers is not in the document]
alemagio: 1- tsc compiles the project into the dist directory (according to tsconfig.json)
2- concurrently allows 'tsc -w' and 'fastify start' (sorry i didn't comment it but when you see '../cli.js' should be 'fastify start') to run together
3- -n "TypeScript,App" -c "yellow.bold,cyan.bold" is for colors in logging

If there is something which is not clear just let me know.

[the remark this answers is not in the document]
mcollina: ok! then make it "fastify-start"

alemagio:
Hi
After a long time code is updated.
Let me know what you think.

templates/app-ts/test/plugins/support.test.ts
```diff
@@ -0,0 +1,11 @@
+import { test } from 'tap'
+const Fastify = require('fastify')
```
alemagio: this is the only one left with the require statement
I could not find a better way to make it work honestly :sweat_smile:

mcollina: @Ethan-Arrowood can you chime in?

Ethan-Arrowood: This is still on Fastify v2 -- try using `import * as Fastify from 'fastify'` and see if you can get it to work with that. You may need to define your server with `const fastify = Fastify.fastify()` or something like that though.

FYI all this is fixed in v3 (which is coming very soon)!

alemagio: Actually i found out i needed to import the extended types i created for the custom plugin:
`import Fastify from 'fastify'`
and 
`import types from '../../@types'`

alemagio:
Code is updated and rebased

templates/app-ts/@types/index.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+declare module 'fastify' {
+  export interface FastifyInstance<
+  HttpServer = http.Server,
+  HttpRequest = http.IncomingMessage,
+  HttpResponse = http.ServerResponse
+> {
+    someSupport(): string;
+  }
+}
```
mcollina: What is this file for?

templates/app-ts/__gitignore
```diff
@@ -0,0 +1,62 @@
...
+*code-workspace
+
+# clinic
+profile*
+*clinic*
+*flamegraph*
+
+# generated code
+examples/typescript-server.js
+test/types/index.js
```
mcollina: These should be wildcards

package.json
```diff
@@ -58,6 +58,8 @@
     "yargs-parser": "^18.1.1"
   },
   "devDependencies": {
```
mcollina: We run the template tests as well. Could you add them to the unit script?

templates/app-ts/services/root.ts
```diff
@@ -0,0 +1,17 @@
+import { FastifyInstance, FastifyRequest, FastifyReply } from "fastify"
+
+module.exports = function (fastify: FastifyInstance, opts: any, next: Function) {
+  fastify.get('/', function (request: FastifyRequest, reply: FastifyReply<any>) {
+    reply.send({ root: true })
+  })
+
+  next()
+}
```
mcollina: I would use the same structure of root.ts

alemagio: I guess you mean root.js :sweat_smile: 
Sure i'll do it

templates/app-ts/@types/index.d.ts
alemagio: Is the extension of the FastifyInstance in order to add the custom plugin to it.
Without it TS will throw an error because the type FastifyInstance has no property called someSupport.

templates/app-ts/__gitignore
alemagio: Sorry what do you mean?

templates/app-ts/@types/index.d.ts
mcollina: Is this customary to put it in a @types folder? can it leave in the same file of the plugin?

templates/app-ts/__gitignore
mcollina: I think in a ts project the js files should be ingnored, right? So, you need a wildcard to match them.

alemagio:
Hi, I don't why some checks in CI succeded and some failed. 
I checked the report and the error looks to be in the new unit script I wrote.
But I don't understand why half of them are working fine.
Let me know if there is something I can do.

mcollina:
Those tests runs on Windows.

Please remove the TS_NODE_PROJECT env variable from the test s or use cross-env, that should work.

Why is that env variable needed?

alemagio:
That variable tells tap where the tsconfig file is when running .ts tests.
Another option could be to create a tsconfig.test.json file for example.
I didn't find another way to do it, I'll try with crossenv but if you have any suggestion to run ts tests I'm open to it.

templates/app-ts/@types/index.d.ts
Ethan-Arrowood: It is by preference. My preference is that types are inlined as thats the most definite way for them to be included in a TS project. This method is valid, but requires users to configure their project to look for them in this directory.

alemagio:
Hi,
code is updated and conflicts are 
@Ethan-Arrowood in this version type declaration is in the plugin file (related comment [here](https://github.com/fastify/fastify-cli/pull/227#discussion_r399812918)

PR #232, opened 2020-03-20
chore(workflows): add GitHub workflows

JamesSingleton:
This takes care of #231, if any naming needs to change let me know.

#### Checklist


JamesSingleton:
An example of it running since GitHub actions don't work in PRs from forks. https://github.com/JamesSingleton/fastify-cli/actions/runs/59449736

.github/workflows/test.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Setup Node
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node }}
+    - name: Install Dependencies
+      run: npm install
+    - name: Run Tests
+      run: npm run test
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mcollina: Is this needed?

JamesSingleton: Yes it is needed to run, but it is created automatically by GitHub.

Eomm: Are our test requiring the `GITHUB_TOKEN` env?
I check rapidly and it doesn't for me, could you tell us why it is required?

JamesSingleton: @Eomm, https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret

mcollina: The question both @Eomm and myself have is why it's needed to run `npm test`. As you see in https://github.com/fastify/fastify/blob/b546ea8aa9335c1816714e03e004f789935f5390/.github/workflows/ci.yml#L42-L46, coveralls action need it. I don't think `npm test` needs it for the tests to pass.

Eomm: Sorry to be "crazy", but if it doesn't need, I would not add it since that token could be read at runtime by our dep and I would do this only with caution in mind

mcollina:
Could you update the badge as well?

JamesSingleton:
> Could you update the badge as well?

https://github.com/fastify/fastify-cli/pull/232/commits/7e404851366c4f11329750bdcbaddc463eec46c0

PR #242, opened 2020-05-06
feat: pass custom plugin options from cli

### LINKED ISSUE #240: Custom plugin options
## 🚀 Feature Proposal
I think it could be useful to allow passing custom options to the plugin defined by the `app.js`. 
After reading the discussion at https://github.com/fastify/fastify-cli/issues/26 I would like to propose an alternative approach: exposing a `pluginOptions` property in `app.js`, the same way the custom options for Fastify are passed using the `options` property.

## Motivation

In some cases, you might need to define custom options for the `app.js` and you might want to be able to pass them directly through the `options` object or let `fastify-cli` inject them.

## Example

**app.js**
```js
module.exports = function (fastify, options, next) {
  ...
  next()
})

module.exports.options = {} // Cusom Fastify options

module.exports.pluginOptions = {} //Custom plugins options injected by fastify-cli
```

dnlup:
Allow passing custom options for the app plugin using additional arguments from the cli. All the args passed after `--` will be parsed and passed to the plugin as options.

Fixes #240 .

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ module.exports = async function (fastify, options) {
 
 For a list of available flags for `fastify start` see the help: `fastify help start`.
 
-If you want to use custom options, just export an options object with your route and run the cli command with the `--options` flag.
+If you want to use custom options for the server creation, just export an options object with your route and run the cli command with the `--options` flag.
 
 ```js
 // plugin.js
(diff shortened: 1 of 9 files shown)


mcollina:
why is this still a draft?

mcollina:
oh it's missing docs! good work on the code!

dnlup:
`args.js` and `start.js` are not tested when running `npm test`. Should I add them?

mcollina:
what do you mean?

dnlup:
I was expecting tests to fail with my changes, but they didn't. Looking at the [unit](https://github.com/fastify/fastify-cli/blob/master/package.json#L11) script it seems that `test/start.test.js` and `test/args.test.js`, which are the one that I was expecting to fail, are not included. I tested them manually, and I was wondering if it was the case to add them, but there's also must be a reason that I am missing for them not being there, I think.

mcollina:
Definitely a mistake. Add them

dnlup:
I added all the files from the `test` folder.

PR #247, opened 2020-05-20
fastify 3.x.x cjs boilerplate poc

davidmarkclements:
#### Checklist

This is a rethinking of the internal architecture for fastify apps.

Broad strokes:

* services folder renamed to routes
* routes and plugins are dealt with in isolation, libs and tests go in same folders
* removing comments, this is a strawman to trigger debate. I think we should provide links to documentation instead to reduce boilerplate/educational noise
* no more `test/helper.js`, created another strawman repo to replace it here: https://github.com/fastify/fastify-tap , this adds `t.fastify` to node-tap, which starts instances automatically AND also provides true plugin isolation (the current approach `helper.js` does not isolate, it starts the full app and then registers the plugin)

Unrelated to this, core cli tests (start.js) are broken, they were passing before because they were removed from package.json and then placed back in this commit: https://github.com/fastify/fastify-cli/commit/0fb5a50455758d2a0ad8d93469e4290537004e27

I've removed them again to avoid the noise of unrelated failing tests


davidmarkclements:
CI fail looks to be infra failure, 

image


choking on gh repo as dep (which is temporary for purposes of this poc)

mcollina:
I have a strong feeling that https://github.com/fastify/fastify-tap/blob/d985f8d4de8001ee7ccfd54c118ec8d3252c9b5f/index.js#L6-L14 would not work for esm.. I would prefer a solution that  is explicit and it does not add things to prototypes. 

I would prefer if we used https://node-tap.org/docs/api/advanced/#taddassertname-length-fn, and potentially make the require of the root app explicit.

davidmarkclements:
fastify-tap updated to not use the approach that may conflict with esm,

The new interface involves passing fastify and tap to fastify-tap to init, loading app.js in the test and passing it to `t.fastify`.

https://github.com/fastify/fastify-cli/pull/247/files#diff-9c14c9b776f9f9fb00622bcb0e7ac41aR4

The addAssert interface must return an assertion, but we want to return the fastify instance, so addAssert doesn't work here. 

@mcollina and I discussed this earlier, the result is to currently leave it as a prototype extension.

However with the new interface we would return a `proxy (new Proxy(tap, ...))` that over rides the `test` function (and also `only`) in order to inject `t.fastify` instead. This would mean that require('tap').test would be different to `require('fastify-tap')(tap, fastify).test` whereas currently they are the same (because we modify `tap.Test.prototype`). While it's definitely not an ideal approach, by exposing the `Test` constructor `tap` and referencing `tap.Test.prototype` in the docs the `tap` project seems to be encouraging this approaching for non-assert extensions.

mcollina:
I don't understand why you need a proxy for the fastify instance

davidmarkclements:
The reason I proxied it is to make it awaitable/thenable as a shorthand for .ready(). The idea is to think of t.fastify as cli start - one operation. 

There are other ways, but proxy makes for seamless inspection (if you console log you’re looking at the instance unmodified). Since it doesn’t modify the instance that also means no side effects other than the intended awaitability. Since it’s for testing performance isn’t a problem. From a programming philosophy perspective this is a meta programming use case which proxy is ideal for, imo. It’s also simpler to cover every thenable/awaitable combination with this approach (all then/catch combos). Finally there’s potential future opportunities for extra testing diagnostic helpers using the existing proxy wrapper

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "standard",
-    "unit": "cross-env TS_NODE_PROJECT=./templates/app-ts/tsconfig.json tap test/*.test.js templates/app/test/*/*.test.js templates/app-ts/test/*/*.test.ts",
+    "unit": "cross-env TS_NODE_PROJECT=./templates/app-ts/tsconfig.json FASTIFY_BOOT_PATH=./templates/app/app.js tap -R classic -j1 --no-esm test/generate-typescript.test.js templates/app/plugins/*/test/*.test.js templates/app/routes/*/test/*.test.js templates/app-ts/test/*/*.test.ts",
```
mcollina, line 11: why is `FASTIFY_BOOT_PATH` needed?
what is `-R classic`?
Please copy the `.taprc` from https://github.com/fastify/fastify/blob/master/.taprc if you want to make tests faster.

Also, why `-j 1`? Parallel testing should make things a bit faster.

templates/app/plugins/README.md
```diff
@@ -1,16 +0,0 @@
...
-Files in this folder are typically defined through the
-[`fastify-plugin`](https://github.com/fastify/fastify-plugin) module,
-making them non-encapsulated. They can define decorators and set hooks
-that will then be used in the rest of your application.
-
-Check out:
-
-* [The hitchhiker's guide to plugins](https://github.com/fastify/fastify/blob/master/docs/Plugins-Guide.md)
-* [Fastify decorators](https://www.fastify.io/docs/latest/Decorators/).
-* [Fastify lifecycle](https://www.fastify.io/docs/latest/Lifecycle/).
```
mcollina, line 16: The documentation of this README file should be put somewhere in the template.

templates/app/plugins/support/test/support.test.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+const fastify = require('fastify')
+const tap = require('tap')
+const { test } = require('fastify-tap')(fastify, tap)
+const support = require('..')
+
+test('support works standalone', async ({ equal, fastify }) => {
+  const plugin = await fastify(support)
```
mcollina, line 8: I don't get this `await`. It's unintuitive and it should not be needed.

mcollina:
CI is failing :(

package.json
davidmarkclements: `FASTIFY_BOOT_PATH` is not longer relevant, I'll remove

`-R classic` makes newer tap use the same reporter we're used to, it's an automatic reflex for me to use it

`-j1` also re-enables streaming (so the tests progress updates as tests complete). Parallel in tap now forces batching, which creates laggy output and degrades devex imo. 

Will do re .taprc but I propose we add -R classic, and possibly -j1 as well.

templates/app/plugins/README.md
davidmarkclements: can we add documentation to this repo, and link to it instead?

templates/app/plugins/support/test/support.test.js
davidmarkclements: It's short hand for:

```
const plugin = fastify(support)
await fastify.ready()
```

Why do you think it's unintuitive?

davidmarkclements: update - caught up on phone with matteo, this isn't needed, will be updating accordingly

templates/app/plugins/README.md
davidmarkclements: caught up on phone - @mcollina is fine with linking docs

package.json
davidmarkclements: caught up on phone, -R classic is going into taprc, -j1 isn't

davidmarkclements:
> CI is failing :(

This will be addressed when general structure is validated. On that can you please weigh in on the new directory structure for plugins/routes (e.g. tests are within plugins/routes instead of by the side of them)

davidmarkclements:
* proxy remove from fastify-tap
* fastify-tap renamed to fastify-test
* boilerplate updates
* .taprc file added to project root and template/app folders

templates/app/routes/_testing/index.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+const { join } = require('path')
+const autoload = require('fastify-autoload')
+
+// use this file to create a plugin that provides reusable state to all routes
+// this is useful for isolating routes by decoupling from root app.js and for
+// creating reusable resource pools, such as a database connection
```
mcollina, line 7: I'm not really ok with this approach. This will lead to global state being managed in a plugin, which defeats the arguments for plugins.

IMHO, fastify-test should support some ways of:

1. configuring the options for the plugins to pass in a connection pool
2. a way to create said  connection pool before the first test is executed and destroyed after the last.

davidmarkclements:
@mcollina I've pushed a strawman config for discussion before I implement the functionality in fastify-test (so tests will fail currently). Lmk what you think

davidmarkclements:
after much discussion with @mcollina over the phone we've come to the conclusion that taking this to a point of full route/plugin testing isolation will need to fundamental adjustments to fastify core - the full story for boilerplate will be more of a v4 thing.

However for v3 I've reduced down the changes to a managable mvp:

* services -> routes
* async functions
* both app and app-ts


@mcollina @delvedor lmk if this works for you

PR #236, opened 2020-03-27
lock pino-colada

Eomm:
We should lock the dep of pino-colada to fix the `-P` parameter
There was already the v2.0.0 on master but it drops the older node so we could not release it with fastify v2

Fixes #235 
Ref https://github.com/lrlna/pino-colada/issues/34

PR #250, opened 2020-06-26
Fix pretty print of logs with -P option

lundibundi:
Basically same as https://github.com/fastify/fastify-cli/pull/236 using fixed pino-colada version

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "help-me": "^1.1.0",
     "is-docker": "^2.0.0",
     "make-promises-safe": "^5.0.0",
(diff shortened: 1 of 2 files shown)


lundibundi:
I tried to fix `fastify-cli` to use `pino-colada@2` but https://github.com/lrlna/pino-colada/pull/29 changed the `parse` function in a strange way so now it will not prettify output of just stdout due to `if (!parsedData.value || parsedData.err || !isPinoLog(parsedData.value)) {` check. Not sure what would be the best way to fix that but reverting to `1.6.1` works fine.

PR #251, opened 2020-06-27
pino colada bug fix

davidmarkclements:
#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "help-me": "^1.1.0",
     "is-docker": "^2.0.0",
     "make-promises-safe": "^5.0.0",
-    "pino-colada": "^1.6.1",
+    "pino-colada": "^2.0.0",
     "pump": "^3.0.0",
     "resolve-from": "^5.0.0",
     "split2": "^3.1.1",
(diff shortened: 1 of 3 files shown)


davidmarkclements:
(also a template fix)

package.json
```diff
@@ -57,6 +57,7 @@
     "pino-colada": "^1.6.1",
```
mcollina, line 57: Can you bump pino-colada to v2.0.0?

mcollina:
Also check out https://github.com/fastify/fastify-cli/pull/250.

cc @lundibundi

lundibundi:
Did you verify this on an actual server output? I wrote seemingly the exact same code and while it did not crash it didn't prettify the output either. (Using `pino-colada@2`)

package.json
davidmarkclements: done

davidmarkclements:
> Did you verify this on an actual server output? I wrote seemingly the exact same code and while it did not crash it didn't prettify the output either. (Using `pino-colada@2`)

good spot, I think pino-coloda is broken:

image

davidmarkclements:
here's the problem:

https://github.com/lrlna/pino-colada/blob/master/index.js#L26

we don't have `v` in latest pino

davidmarkclements:
I've PR'd pino colada 

https://github.com/lrlna/pino-colada/pull/35

davidmarkclements:
updated for latest pino-coloda,I think we're good to merge

PR #262, opened 2020-08-16
Improve ts template and bump fastify-plugin

lundibundi:
#### Checklist

This is not pretty but I don't see a way around the way tap typings are defined.
`export type Test = typeof tap['Test']['prototype'];`
Previously `Test` type would just be any due to `noImplicitAny: false` by default.

* use strict: true in tsconfig.json
* include tsc in `npm test` to check types before running tests
* use ES2018 as TS target
* update code to consistently use single quotes and have capitalized
  comments
* add @types/tap and fix typings for tests
* change exports for plugins and routes to use correct types with
  type inference
* consistently use/ignore promises returned by fastify functions
* move all source to src/ folder
* create separate tsc target for test folder and run it before tap

/cc @Ethan-Arrowood


Ethan-Arrowood:
Quick glance looks great! Will try and review sometime in the next few days

templates/app-ts/app.ts
```diff
@@ -1,25 +1,34 @@
...
+): Promise<void> => {
   // Place here your custom code!
 
   // Do not touch the following lines
 
   // This loads all plugins defined in plugins
   // those should be support plugins that are reused
   // through your application
-  fastify.register(AutoLoad, {
+  await fastify.register(AutoLoad, {
```
mcollina, line 20: This await is not really needed right now.

templates/app-ts/app.ts
```diff
@@ -1,25 +1,34 @@
...
     dir: path.join(__dirname, 'plugins'),
     options: opts
   })
 
   // This loads all plugins defined in routes
   // define your routes in one of these
-  fastify.register(AutoLoad, {
+  await fastify.register(AutoLoad, {
```
mcollina, line 27: also this

templates/app-ts/app.ts
lundibundi: Not really familiar with the internal flow of this so I may misunderstand. 
As I understand it:
should it be "plugin return promise that is resolved when it is done loading/doing everything it needs" rather than
"plugin returns a promise that is resolved when everything is scheduled to be loaded"?

mcollina: It does not return a promise. The server is a thenable, and by doing an `await` all the plugins registered so far are loaded. The same happens after the plugin function resolves. The only reason somebody would add this is if they want to use a decorator _inside_ the plugin function but this is not the case.

Note that keeping them does not add any problem, however I find it confusing

templates/app-ts/tsconfig.json
```diff
@@ -1,22 +1,22 @@
...
-            ]
+            "*": ["node_modules/*"]
         }
     },
     "include": [
+        "app.ts",
         "routes/**/*.ts",
         "plugins/**/*.ts",
-        "app.ts"
+        "test/**/*.ts"
```
fox1t, line 20: We don't need to include test files in the tsconfig.json that handles build for our app.

lundibundi:
> Since we are improving this, it would be better to put all .ts files inside an src folder.
The test however should remain in test folder. Another nice addition would be to have a separate tsconfig.json file inside test folder that extends the top level one and that has "include": prop set to . (test folder itself)

I initially wanted to introduce `src` folder but thought it would be too "breaking" for this, would gladly change it if there is consensus.

What is the need for separate test `tsconfig` though, wouldn't it be just the same (empty with extends and `.`(test) folder)?

fox1t:
There is no need to build test code when running `npm run build` to build the app. Am I missing something?

lundibundi:
@fox1t As I see it - if tests fail to build that means something is wrong with the code/api and has to be changed in one place or the other. Like tests but on a type level. Therefore they have to be a part of the basic build.
Though I understand that this may not be a general opinion so it's okay to move them to a separate config/task if that is needed.

fox1t:
I think that we need to see test in the TS as we see them in JS: we run them only when we want to test. Same has to be in TS: compile and run test files only when testing. In addition to that there is no need to ship compiled tests code inside dist folder.

templates/app-ts/app.ts
lundibundi: Oh, now I understand. Only if you depend on the modifications from other plugins then you would have to wait for them to load.
Changed these to `void fastify.register...` to pacify some very strict ESLint rules (no-floating-promises).

templates/app-ts/tsconfig.json
lundibundi: Done

templates/app-ts/src/app.ts
```diff
@@ -1,25 +1,34 @@
-import AutoLoad from 'fastify-autoload';
 import * as path from 'path';
```
fox1t, line 1: `esModuleInterop` is set to true, no need to use star import.

templates/app-ts/tsconfig.json
```diff
@@ -1,22 +1,17 @@
...
-        "moduleResolution": "node",
+        "strict": true,
         "sourceMap": true,
         "outDir": "dist",
         "baseUrl": ".",
         "paths": {
-            "*": [
-                "node_modules/*"
-            ]
+            "*": ["node_modules/*"]
```
fox1t, line 13: This is not really needed since node_modules is already included.

fox1t (changes requested):
Nice job! Thanks for this PR.
There are some minor things to address and then we can merge this.

templates/app-ts/tsconfig.json
lundibundi: I think this will be needed for `tsconfig.test.json`, I'll move it there (with a `../`).

fox1t: As far as I know, `node_modules` is always included when `moduleResolution` is set to `node`, otherwise if multiple levels of`node_modules` are provided it will not work.
More info here: https://www.typescriptlang.org/docs/handbook/module-resolution.html#node

lundibundi: Oh, yeah, I forgot about walking "up" the hierarchy. Then this can just be removed.

mcollina (approved):
lgtm

lundibundi:
ping @Ethan-Arrowood @fox1t all outstanding comments should be fixed now.

generate.js
```diff
@@ -57,6 +57,7 @@ const typescriptTemplate = {
   },
   devDependencies: {
     '@types/node': cliPkg.devDependencies['@types/node'],
+    '@types/tap': cliPkg.devDependencies['@types/node'],
```
Ethan-Arrowood, line 60: Is this meant to be @types/tap in both places?

templates/app-ts/tsconfig.json
```diff
@@ -1,22 +1,14 @@
 {
     "compilerOptions": {
-        "module": "commonjs",
+        "target": "ES2018",
+        "module": "CommonJS",
```
fox1t, line 4: In the fastify organization is used "commonjs" almost everywhere: it would be nice to have it also here. Both versions are schema compliant, though.

fox1t (approved):
LGTM

templates/app-ts/tsconfig.json
lundibundi: I'll then change this for consistency with other repos.

fox1t (approved):
LGTM

lundibundi:
ping @fox1t @mcollina @Ethan-Arrowood I think this can be landed. =)

Issue #258, opened 2020-08-07
fastify routes command is broken with v3 applications


tchak:
## 🐛 Bug Report

`fastify routes` command is broken with v3 applications

## To Reproduce

```bash
fastify routes
```

## Stacktrace

```
TypeError: printRoutes(...).close is not a function
```

## Your Environment

- *node version*: 12
- *fastify version*: >=3.0.0
- *os*: Mac

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #265, opened 2020-09-02
Fix fastify routes command

lundibundi:
Fixes: https://github.com/fastify/fastify-cli/issues/258

#### Checklist

I've added a test though it may be flaky so I'm not sure we want it.

### DIFF
--- a/test/print-routes.test.js
+++ b/test/print-routes.test.js
@@ -22,6 +22,21 @@ test('should print routes', async t => {
   t.ok(spy.calledWithMatch('debug', '└── / (GET|POST)\n'))
 })
 
+test('should print routes via cli', t => {
+  t.plan(2)
+
+  const spy = sinon.spy()
+  const command = proxyquire('../print-routes', {
+    './log': spy
+  })
+  command.cli(['./examples/plugin.js'])
+
+  setTimeout(() => {
+    t.ok(spy.called)
+    t.ok(spy.calledWithMatch('debug', '└── / (GET|POST)\n'))
+  }, 5000)
+})
+
 test('should warn on file not found', t => {
   t.plan(1)
 
--- a/test/print-routes.test.js
+++ b/test/print-routes.test.js
@@ -22,19 +22,17 @@ test('should print routes', async t => {
   t.ok(spy.calledWithMatch('debug', '└── / (GET|POST)\n'))
 })
 
-test('should print routes via cli', t => {
+test('should print routes via cli', async t => {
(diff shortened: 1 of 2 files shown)


fox1t (approved):
LGTM

test/print-routes.test.js
```diff
@@ -22,6 +22,21 @@ test('should print routes', async t => {
...
 })
 
+test('should print routes via cli', t => {
+  t.plan(2)
+
+  const spy = sinon.spy()
+  const command = proxyquire('../print-routes', {
+    './log': spy
+  })
+  command.cli(['./examples/plugin.js'])
```
Eomm, line 32: Could you add a test with the `--help`?

Reading the code, this case should not return a promise, so the `.then` should break.
So maybe it is necessary to move the `if(fastify)` in the upperscope

lundibundi: I'll look in detail later but from what I've seen the `help` case uses explicit `process.exit` so that `.then` will never execute.

Eomm: You are right!

https://github.com/fastify/fastify-cli/blob/056c0b510b27d886ddb9c68be63fca42fc37b935/util.js#L12

(in this case, the `if` statement should not be necessary)

test/print-routes.test.js
```diff
@@ -22,6 +22,21 @@ test('should print routes', async t => {
...
+
+  setTimeout(() => {
+    t.ok(spy.called)
+    t.ok(spy.calledWithMatch('debug', '└── / (GET|POST)\n'))
+  }, 5000)
```
mcollina, line 37: why is this needed? can you please use fake timers? Or maybe refactor to avoid needing this in the first place?

lundibundi: The `cli` doesn't return a promise or any other indication that it has completely finished so I opted in to just wait a bit.
I'm not sure what would be the best approach here. I could make the `cli` return a promise but that's imo not a very good solution. We could also use `process on exit` though I'm not sure how `tap` works with that. I'm not sure how I can use fake timers to handle that though.

Also, as I mentioned in a PR comment we could just avoid this test.

mcollina: Let's make the cli return a promise instead.

lundibundi: Done

lundibundi:
ping @fox1t @mcollina @Ethan-Arrowood I think this can be landed. =)

PR #269, opened 2020-09-11
Generate plugin

alemagio:
#### Checklist


README.md
```diff
@@ -204,14 +204,28 @@ It is a standard Fastify plugin and you will not need to add the `listen` method
...
+### generate-plugin
+
+`fastify-cli` can help you improve your plugin development by generating a scaffolding project:
+
+1. `fastify generate <yourplugin>`
+2. `cd yourplugin`
+3. `npm install`
+
+Your boilerplate will provide you some useful npm scripts:
+* `npm unit`: runs all unit tests
```
zekth, line 225:
```suggestion
* `npm run unit`: runs all unit tests
```

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+
+const pluginTemplate = {
+  dir: 'plugin',
+  main: 'index.js',
+  types: 'index.d.ts',
+  scripts: {
+    lint: 'standard && npm run lint:typescript',
+    'lint:typescript': 'standard --parser @typescript-eslint/parser --plugin @typescript-eslint/eslint-plugin test/types/*.ts',
+    test: 'npm run lint && npm run unit && npm run typescript',
+    typescript: 'tsd',
```
zekth, line 24:
```suggestion
    test:typescript: 'tsd',
```
to follow fastify semantic: https://github.com/fastify/fastify/blob/578602f424314ae9d78cea68cbc2b419b8daec51/package.json#L15

README.md
```diff
@@ -204,14 +204,28 @@ It is a standard Fastify plugin and you will not need to add the `listen` method
...
+* `npm lint`: to check your project's code style
```
zekth, line 226: same comment as unit

README.md
```diff
@@ -204,14 +204,28 @@ It is a standard Fastify plugin and you will not need to add the `listen` method
...
+* `npm typescript`: runs types tests
```
zekth, line 227: same comment as unit

README.md
```diff
@@ -204,14 +204,28 @@ It is a standard Fastify plugin and you will not need to add the `listen` method
...
+* `npm test`: runs all the checks at once
```
zekth, line 228: same comment as unit

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+            return reject(err)
+          }
+
+          template.logInstructions(pkg)
+          resolve()
+        })
+      })
+    })
+  })
+}
```
mcollina, line 97: Can you use async/await instead? You can convert callback-based functions to promise equivalent with `util.promisify`.

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+    typescript: cliPkg.devDependencies.typescript
+  },
+  tsd: {
+    directory: 'test'
+  },
+  logInstructions: function (pkg) {
+    log('debug', 'saved package.json')
+    log('info', `project ${pkg.name} generated successfully`)
+    log('debug', `run '${chalk.bold('npm install')}' to install the dependencies`)
+    log('debug', `run '${chalk.bold('npm test')}' to execute the tests`)
```
zekth, line 47:
```suggestion
    log('debug', `run '${chalk.bold('npm run test')}' to execute the tests`)
```

zekth (changes requested):
Some minor comments

README.md
alemagio: the test command is `npm test` in other parts of the readme, I think npm should be able to run it even without `run` keyword.
What do you think?

zekth: Right, `test` works. The npm `cli` is a bit inconsistent on this point.

alemagio: True

fox1t: `test` is considered to be at the same level as install and other top levels. This goes back to the early days of npm :)

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+const { execSync } = require('child_process')
+const log = require('./log')
+
+const pluginTemplate = {
+  dir: 'plugin',
+  main: 'index.js',
+  types: 'index.d.ts',
+  scripts: {
+    lint: 'standard && npm run lint:typescript',
+    'lint:typescript': 'standard --parser @typescript-eslint/parser --plugin @typescript-eslint/eslint-plugin test/types/*.ts',
```
fox1t, line 22: Shall we put these configurations in a dedicated `.eslintrc` file?

generate-plugin.js
fox1t: I would prefer to leave it as `npm test` since it is supported and widely use like this.

templates/plugin/.taprc
```diff
@@ -0,0 +1,4 @@
+esm: false
```
fox1t, line 1: We can use the same config as it is here: https://github.com/fastify/fast-json-stringify/blob/master/.taprc

Why setting jobs to 1?

fox1t (changes requested):

README.md
alemagio: TIL, thanks for the explanation.

templates/plugin/.taprc
alemagio: I simply used the same `.taprc` of the [fastify-cli](https://github.com/fastify/fastify-cli/blob/master/.taprc).
But maybe it was not a good idea since the fastify-cli is a bit different from a simple plugin.
I'll take your suggestion.

generate-plugin.js
alemagio: After further investigation I found out that `standard` has some issues with TS even using the parser and the plugin.
I found [ts-standard](https://github.com/toddbluhm/ts-standard) which looks good and needs no config but works out of the box.
This way I could also remove those dependencies and we need no more configuration.
What do you think?

mcollina (approved):
lgtm

alemagio:
@fox1t @zekth I think I solved all your comments, still missing something?

zekth (approved):
LGTM

templates/plugin/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "esModuleInterop": true,
+        "allowSyntheticDefaultImports": true,
```
fox1t, line 5: This is not needed since `esModuleInterop` already set it to true.

templates/plugin/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+        "target": "es6",
+        "moduleResolution": "node",
+        "sourceMap": true,
+        "outDir": "dist",
+        "baseUrl": ".",
```
fox1t, line 10: Since we are targeting Node.js and not an AMD I would completely drop both, `baseUrl` and `paths` since they could create "bizzare" issues, hard to debug, and should increase the rate of issues on fastify repos that are related to TS. You can find more info here: https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url

fox1t (changes requested):

fox1t:
@alemagio I've checked it out. Sorry for being so pedantic but we have plenty of issues opened across all of our repositories just because of wrong `tsconfig` files used by our users. I really want to be sure to not put into the wild potentially devastating configurations.

I am working on shared fastify tsconfig that you can extend from. It would be nice to use it here and change only the relevant part. Doing this we will be sure what we need to look for when a TS compilation time error pop-ups.

alemagio:
> @alemagio I've checked it out. Sorry for being so pedantic but we have plenty of issues opened across all of our repositories just because of wrong `tsconfig` files used by our users. I really want to be sure to not put into the wild potentially devastating configurations.
> 
> I am working on shared fastify tsconfig that you can extend from. It would be nice to use it here and change only the relevant part. Doing this we will be sure what we need to look for when a TS compilation time error pop-ups.
> 
> You can find a preview here: https://github.com/fastify/tsconfig
> 
> As soon as it gets published on npmjs we could update this PR.

@fox1t you have to be pedantic, I'm here to learn first of all so any comment or suggestion is well accepted.
I'll push an update asap. :smile:

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+  const generifyPromise = promisify(generify)
+  try {
+    const file = await generifyPromise(
+      path.join(__dirname, 'templates', template.dir),
+      dir,
+      {}
+    )
+    log('debug', `generated ${file}`)
+  } catch (err) {
+    throw new Error(err)
```
RafaelGSS, line 61: This try/catch is not necessary if you will throw the same error only in a different stack trace. I would treat the error with a better description.

generate-plugin.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+  process.chdir(dir)
+  execSync('npm init -y')
+
+  log('info', `reading package.json in ${dir}`)
+  let data
+  try {
+    data = await readFile('package.json')
```
RafaelGSS, line 69: On this case, I would use something like:

```suggestion
    pkg = await readFile('package.json').then(JSON.parse)
```
or
```suggestion
    packageString = await readFile('package.json')
    pkg = JSON.parse(packageString)
```

templates/plugin/tsconfig.json
RafaelGSS: @fox1t is not better to wait for [fastify-tsconfig](https://github.com/fastify/tsconfig) land and use here?

alemagio:
> @alemagio I've checked it out. Sorry for being so pedantic but we have plenty of issues opened across all of our repositories just because of wrong `tsconfig` files used by our users. I really want to be sure to not put into the wild potentially devastating configurations.
> 
> I am working on shared fastify tsconfig that you can extend from. It would be nice to use it here and change only the relevant part. Doing this we will be sure what we need to look for when a TS compilation time error pop-ups.
> 
> You can find a preview here: https://github.com/fastify/tsconfig
> 
> As soon as it gets published on npmjs we could update this PR.

@fox1t I think, we could, as a momentary solution, set the ts-config as it is on the shared ts-config (with maybe a few changes, like outDir) and open a new issue on this repo to update it and make it extend the shared one as soon as it is released.
What do you think?

fox1t:
@alemagio I am going to release the tsconfig today! I think we can wait for a couple of hours and use it here.

alemagio:
> @alemagio I am going to release the tsconfig today! I think we can wait for a couple of hours and use it here.

Sounds great!

fox1t:
You can use: https://www.npmjs.com/package/fastify-tsconfig :)

alemagio:
Perfect, thank you so much.

alemagio:
@RafaelGSS I removed the try/catch and some others since, you are right throwing the same error makes no sense.
I did not add any king of logging since at that point, I believe, it is better to just let each part log its own error/stacktrace.
Or at least I could not figure out something meaningful to log for the user.
What do you think?

templates/plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "fastify-tsconfig",
+    "compilerOptions": {
+      "outDir": "dist"
+    },
+    "include": [
```
fox1t, line 6: We don't need this include since this is a template for JS plugin that has also TS typings.

templates/plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "fastify-tsconfig",
+    "compilerOptions": {
+      "outDir": "dist"
```
fox1t, line 4: We don't need this outDir since this is a template for JS plugin that has also TS typings.

fox1t (changes requested):

fox1t:
Awesome work! I am still a little bit confused about some TS settings: this is a template for JS plugin that has also a toolchain for shipping TS typings, right?

alemagio:
Exactly.
I actually added ts-config file just because ts-standard needs it.

templates/plugin/tsconfig.json
fox1t: I suppose we can leave this for eslint and standard.

templates/plugin/tsconfig.json
fox1t: We can still remove this outDir and add noEmit since this is needed only for standard

fox1t (changes requested):

RafaelGSS:
> @RafaelGSS I removed the try/catch and some others since, you are right throwing the same error makes no sense.
> I did not add any king of logging since at that point, I believe, it is better to just let each part log its own error/stacktrace.
> Or at least I could not figure out something meaningful to log for the user.
> What do you think?

If all the possibilities that could throw an error are able to be tested throwing an error, that's ok.

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
```
fox1t, line 1: Where is used this configuration?

alemagio: It is necessary to test the app-ts template.
I was planning to add a new issue to use the new shared tsconfig for that template too.

fox1t: ok nice!

fox1t (approved):
LGTM

fox1t:
@alemagio another good addition would be ts-plugin template. I think we should add a new flag `--ts` that genreates app/plugin using TS instead of JS.

alemagio:
@fox1t I was thinking about it and I was planning to make it using the `--lang=ts` option we already have for the app generation.
Work for another saturday morning I guess :wink:

RafaelGSS:
@alemagio On coverage report not shows the test for `generate.js`, there is tested though. Could you include it on the coverage report?

alemagio:
@RafaelGSS Is it possible that it is a master issue too?

RafaelGSS:
I don't know. I mean, the `generate-plugin.test.js` not show any coverage, and the `generate-plugin.js` was added by this PR.

alemagio:
This is from `master`

coverage


This is from this branch 

coverage_generate_plugin

Neither of the two shows `generate.js`, am I wrong?

RafaelGSS:
I mean the `generate-plugin.js`. Is the file that this PR added, right?

alemagio:
@RafaelGSS 
Yes it is.
Maybe I did not understand what you expected.
I think you would expect to see `generate-plugin.js` in the second screenshot (this branch coverage report) correct?
I don't know actually how to make it appear there (I looked on the tap docs a lot but could not figure out a solution) :sweat_smile:.
I also noticed that in the `master` branch (and in this one too) other files are missing from the coverage report, for example `generate.js` and `start.js` and I don't know if this is expected.

RafaelGSS:
@fastify/plugins anyone knows why some branches are not showed on coverage?

RafaelGSS (approved):
I'll create an issue for unexpected behaviour above. LGTM.

PR #272, opened 2020-09-17
Fix watch test

alemagio:
#### Checklist

### DIFF
--- a/lib/watch/fork.js
+++ b/lib/watch/fork.js
@@ -20,9 +20,11 @@ process.on('message', function (event) {
   if (event === GRACEFUL_SHUT) {
     const message = chalk.red('[fastify-cli] process forced end')
     setTimeout(exit.bind({ message }), TIMEOUT).unref()
-    fastify.close(() => {
-      process.exit(0)
-    })
+    if (fastify) {
+      fastify.close(() => {
+        process.exit(0)
+      })
+    }
   }
 })
 
--- a/lib/watch/fork.js
+++ b/lib/watch/fork.js
@@ -24,6 +24,8 @@ process.on('message', function (event) {
       fastify.close(() => {
         process.exit(0)
       })
+    } else {
+      process.exit(0)
     }
   }
 })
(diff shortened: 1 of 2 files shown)


fox1t:
Ok it worked ` test/start.test.js 1> [fastify-cli] process forced end`. I think we can leave this active also on Actions.

fox1t (approved):
LGTM

lib/watch/fork.js
```diff
@@ -20,9 +20,11 @@ process.on('message', function (event) {
...
     const message = chalk.red('[fastify-cli] process forced end')
     setTimeout(exit.bind({ message }), TIMEOUT).unref()
-    fastify.close(() => {
-      process.exit(0)
-    })
+    if (fastify) {
+      fastify.close(() => {
+        process.exit(0)
+      })
+    }
```
mcollina, line 27: I think this is missing an `else` clause in which we exit immediately if `fastify` is not there. There is no point in waiting.

mcollina (changes requested):
lgtm

lib/watch/fork.js
fox1t: Good point! In that case, we don't need a `GRACEFUL_SHUT` at all!

alemagio: by adding
`else {
  process.exit(0)
}`
makes it exit without waiting for the message `[fastify-cli] process forced end` to appear.
Is that ok?

mcollina: yes, the app is not started, there is nothing to wait.

mcollina (approved):
lgtm

alemagio:
Is the CI failure related to this?

mcollina:
Likely, keep skipping that test on CI.

fox1t:
I don't understand why waiting makes it pass also on CI. Any clues?

alemagio:
Probably that's why it was skipped in the first place.

alemagio:
Maybe it is not related to the CI failure but I just noticed that the process should exit with code 1, like it does in case of timeout, right?

mcollina:
Let's skip that test, exiting with 0 is good.

Issue #246, opened 2020-05-20
npm run dev in typescript project, watch is not working


JoHuang:
## 🐛 Bug Report

Changed Typescript files cannot be applied.

## To Reproduce

1. Use master branch to generate typescript project
2. `npm run dev`
3. change string in example.ts ( server auto restarted )
4. /example response didn't change.
5. I have to ctrl+c to terminate `npm run dev`, run agin to make the change applied.
## Expected behavior
/example response changed according to 3.'s change.
## Your Environment

- *node version*: 12
- *fastify version*: 3.0.0-rc2
- *os*: Mac

setup step:
https://github.com/fastify/fastify-cli/issues/131#issuecomment-630192417

fvelcker:
I have a similar issue. I have to make a change a second time to see the first change happening. if I do a third change, I'll see th second, etc.

fvelcker:
Ah well, sorry, found why... (looks like typing my message rubber ducked me).
`fastify start -w` was running before `tsc` finishes compiling because it detected the change in the `.ts` file.

So I added `--ignore-watch='/src'` as my ts files are in a src folder. So now `fastify start -w --ignore-watch='/src'` restarts only after the files have compiled.

Not sure it is related, but if it helps....

alemagio:
I think this is solved in the latest version of the code.

mcollina:
I just shipped that, please check again.

alemagio:
There is an issue indeed.
The problem is that when there is a change on a TS file both tsc and fastify-cli restart.
But, apparently, fastify-cli is faster than tsc so it restarts before tsc has finished transpiling TS to the dist folder so the change is not loaded.
This is an hotfix:
`tsc && concurrently -k -p \"[{name}]\" -n \"TypeScript,App\" -c \"yellow.bold,cyan.bold\"  \"tsc -w\" \"fastify start --ignore-watch=.ts$ -w -l info -P dist/app.js\""`
If no one else want to I'll push a fix on the template script.

mcollina:
Go for that!

alemagio:
Working hard on this.
Hope to open a PR (even a draft maybe) by the end of the week.
Just to let you know still a WIP.

fox1t:
Ok. Just open a WIP PR and we will discuss on the best way to handle this.

alemagio:
the fix is described [here](https://github.com/fastify/fastify-cli/issues/246#issuecomment-700819896) 
[...]

PR #276, opened 2020-10-09
feat: start can now watch ts files (#246)

alemagio:
#### Checklist

The idea is that we should make the user run `fastify start...` on .ts files and watch for changes on those files.
At the moment the `npm run dev` script in ts project template uses `concurrently` so we basically have 2 watchers running together: one for ts files and one for the compiled `dist` files.
This is not ideal.
This approach uses only the tsc API to watch for .ts files changes. Every time it detects a change it kills fastify process, re-compiles changed files, restarts fastify.

Note: for the moment I assumed that if the target file is a .ts file we always want to watch it even without `-w` option.
Note: [docs](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-

Any suggestion is well accepted.

### DIFF
--- a/lib/watch/tsc-watcher.js
+++ b/lib/watch/tsc-watcher.js
@@ -17,7 +17,7 @@ let child
 
 const emitter = new EventEmitter()
 
-async function watchMain (args, opts) {
+function watchMain (args, opts) {
   const configPath = ts.findConfigFile(
     './',
     ts.sys.fileExists,
@@ -50,7 +50,7 @@ async function watchMain (args, opts) {
   const appFile = resolve(configFile.compilerOptions.outDir, basename(opts._[0]).replace(/.ts$/, '.js'))
   args.splice(args.length - 1, 1, appFile)
 
-  host.afterProgramCreate = async program => {
+  host.afterProgramCreate = program => {
     if (child) {
       child.send(GRACEFUL_SHUT)
     }
--- a/lib/watch/tsc-watcher.js
+++ b/lib/watch/tsc-watcher.js
@@ -1,0 +1,104 @@
+'use strict'
+
+const ts = require('typescript')
+const { EventEmitter } = require('events')
+const formatHost = {
+  getCanonicalFileName: path => path,
+  getCurrentDirectory: ts.sys.getCurrentDirectory,
+  getNewLine: () => ts.sys.newLine
+}
+const { resolve, join, basename } = require('path')
+const cp = require('child_process')
+const forkPath = join(__dirname, './fork.js')
+const {
+  GRACEFUL_SHUT
+} = require('./constants')
+let child
+
+const emitter = new EventEmitter()
+
+async function watchMain (args, opts) {
+  const configPath = ts.findConfigFile(
+    './',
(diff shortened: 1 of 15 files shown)


lib/watch/tsc-watcher.js
```diff
@@ -0,0 +1,104 @@
...
+    return origCreateProgram(rootNames, options, host, oldProgram)
+  }
+  const origPostProgramCreate = host.afterProgramCreate
+
+  const configFile = require(resolve(process.cwd(), configPath))
+
+  const appFile = resolve(configFile.compilerOptions.outDir, basename(opts._[0]).replace(/.ts$/, '.js'))
+  args.splice(args.length - 1, 1, appFile)
+
+  host.afterProgramCreate = async program => {
```
mcollina, line 53: why is this async?

test/start-typescript.test.js
```diff
@@ -0,0 +1,63 @@
...
+      if (err) return reject(err)
+      return resolve({ response, body })
+    })
+  })
+}
+const start = require('../start')
+
+// FIXME
+// paths are relative to the root of the project
+// this can be run only from there
```
mcollina, line 24: what does this mean? Can you create an issue or fix it here?

test/start-typescript.test.js
```diff
@@ -0,0 +1,63 @@
...
+const { once } = require('events')
+const fs = require('fs')
+const crypto = require('crypto')
+const baseFilename = `${__dirname}/fixtures/test_${crypto.randomBytes(16).toString('hex')}`
+
+const t = require('tap')
+const test = t.test
+const sgetOriginal = require('simple-get').concat
+const sget = (opts, cb) => {
+  return new Promise((resolve, reject) => {
```
mcollina, line 13: please use promisify instead.

start.js
```diff
@@ -12,14 +12,16 @@ const pump = require('pump')
 const isDocker = require('is-docker')
 const listenAddressDocker = '0.0.0.0'
 const watch = require('./lib/watch')
+const tsWatch = require('./lib/watch/tsc-watcher')
```
mcollina, line 15: can you load this only on demand? We are loading the entirety of typescript even if it's not a ts project.

mcollina (changes requested):

test/start-typescript.test.js
alemagio: this is present in start.test.js too, sure, I'll fix it.

test/start-typescript.test.js
alemagio: I used promisify.custom because, I think, it is not possible to simply use promisify if the callback has more than one param plus the err, in this case `(err, response, body)`.
Am I wrong?

lib/watch/tsc-watcher.js
alemagio: remaining of experiments
fixed now

mcollina (approved):
lgtm

L2jLiga (approved):
lgtm

fox1t (approved):
LGTM

PR #282, opened 2020-10-18
Ts watch fix

### LINKED ISSUE #279: Typescript handling broken in 2.4
## 💥 Regression Report

#276 (2.4.0) screwed up our whole typescript dev workflow. We had it working perfectly fine (including watching) before with simply doing
```
ts-node ./node_modules/.bin/fastify start -w -l info -P -p 3001 src/index.ts
```
Nothing else was required to get it working, no code changes etc. (Imho this would also be an easier solution than all the code changes in #276)

In the new version fastify-cli now first errors with 
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:122:11)
    at resolve (path.js:980:7)
    at watchMain (/Users/tschinde/Documents/rg/tech-dept-app/api/node_modules/fastify-cli/lib/watch/tsc-watcher.js:50:24)
```
because we do not specify an `outDir` directly in the main typescript config, but rather in the some inherited config. Even after changing this and defining an `outDir` and `rootDir` the emitting of files is not working correctly. I have my source in `src/` and the code emitted is going into `lib/src/` but that is not where it should go, it should just go into `lib/`. So the app crashes because it cannot find files.

my tsconfig (with outDir and rootDir specified to fix above, with ts-node only `rootDir: "./"` was set which is also needed for the typescript tests) is:
```
{
    "include": [
      "src"
    ],
    "extends": "../tsconfig.json",
    "compilerOptions": {
      "outDir": "lib",
      "rootDir": "src"
    }
  }
```

`../tsconfig.json` is:
```
{
  "exclude": ["node_modules", "__tests__", "dist", "lib"],
  "compilerOptions": {
    "lib": ["es2019"],
    "target": "es2019",
    "module": "commonjs",
    "rootDir": ".",
    "outDir": "lib",
    "strict": true,
    "allowJs": false,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "resolveJsonModule": true,
    "baseUrl": ".",
    "declaration": false,
    "sourceMap": false,
    "allowSyntheticDefaultImports": true,
    "moduleResolution": "node",
    "isolatedModules": true
  }
}
```

For now I reverted and will stay on latest 2.3.x

## Last working version

Worked up to version: 2.3.x

Stopped working in version: 2.4.0

## Your Environment

- *node version*: 12
- *fastify version*: 3.6.0
- *os*: Mac
- *ts*: 3.9.7

alemagio:
#### Checklist

This is a proposal to fix #279 
This should fix 2 issues:
1) parent tsconfig are now included so now is possible to specify outDir in parent config
2) outDir is now relative to the config file where it is specified

Note
As you can see there is a failing test, this happens only when running together the tests of the suite `start-typescript.test.js` with error: `child test left in queue: t.test should start the server with watch option in nested dir`
No idea why

I also added a new option `ts-watcher` which can be set to false to allow the user to not watch ts files and to use something like `ts-node` together with fastify-cli watcher.

If we still decide to revert is not a problem off-course but I would still love to have your opinion on this if possible @mcollina and @fox1t (I'm here to learn after all :smile: )

### DIFF
--- a/lib/watch/tsc-watcher.js
+++ b/lib/watch/tsc-watcher.js
@@ -56,9 +56,12 @@ function watchMain (args, opts) {
     }
     origPostProgramCreate(program)
 
+    // We need to set the NODE_PATH variable in case node_modules are in a child directory
+    const env = Object.assign({}, process.env, { NODE_PATH: resolve(process.cwd(), 'node_modules') })
     child = cp.fork(forkPath, args, {
       cwd: process.cwd(),
-      encoding: 'utf8'
+      encoding: 'utf8',
+      env
     })
     let readyEmitted = false
 
--- a/lib/watch/tsc-watcher.js
+++ b/lib/watch/tsc-watcher.js
@@ -45,7 +45,7 @@ function watchMain (args, opts) {
   }
   const origPostProgramCreate = host.afterProgramCreate
 
-  const configFile = require(resolve(process.cwd(), configPath))
+  const configFile = requireTsConfig(resolve(process.cwd(), configPath))
 
   const appFile = join(resolve(dirname(configPath), configFile.compilerOptions.outDir), basename(opts._[0]).replace(/.ts$/, '.js'))
   args.splice(args.length - 1, 1, appFile)
@@ -99,6 +99,30 @@ function reportDiagnostic (diagnostic) {
 
 function reportWatchStatusChanged (diagnostic) {
   console.info(ts.formatDiagnostic(diagnostic, formatHost))
+}
+
+function requireTsConfig (configPath) {
+  const isModule = basename(configPath) === configPath
+  let configFile = require(isModule ? configPath : resolve(process.cwd(), configPath))
+  const parent = configFile.extends
+  if (configFile.compilerOptions.outDir) {
+    configFile.compilerOptions.outDir = resolve(dirname(configPath), configFile.compilerOptions.outDir)
+  }
+  if (parent) {
+    configFile = deepMergeConfig(requireTsConfig(parent), configFile)
+  }
+  return configFile
+}
+
+function deepMergeConfig (obj1, obj2) {
+  const result = Object.assign({}, obj1, obj2)
+  for (const key of Object.keys(result)) {
+    if (typeof result[key] === 'object' && !Array.isArray(result[key])) {
+      result[key] = deepMergeConfig(obj1[key], obj2[key])
+      continue
+    }
+  }
+  return result
 }
 
 module.exports = watchMain
--- a/.gitignore
+++ b/.gitignore
@@ -51,8 +51,9 @@ test.sock
 
 # test artifacts
 test/workdir
-test/fixtures/*.js
-test/fixtures/*.ts
+test/fixtures/**/*.js
+test/fixtures/**/*.ts
+test/fixtures/**/*.json
 .vscode/launch.json
 
 # ts compiled
--- a/args.js
+++ b/args.js
@@ -8,7 +8,7 @@ module.exports = function parseArgs (args) {
       'populate--': true
     },
     number: ['port', 'inspect-port', 'body-limit', 'plugin-timeout'],
-    boolean: ['pretty-logs', 'options', 'watch', 'debug'],
+    boolean: ['pretty-logs', 'options', 'watch', 'debug', 'ts-watcher'],
     string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'tsconfig'],
     envPrefix: 'FASTIFY_',
     alias: {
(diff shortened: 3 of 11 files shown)


mcollina:
@fastify/typescript wdyt?

---

> As you can see there is a failing test, this happens only when running together the tests of the suite start-typescript.test.js with error: child test left in queue: t.test should start the server with watch option in nested dir
No idea why

You need to reduce the test concurrancy to one, see https://node-tap.org/docs/api/parallel-tests/

alemagio:
> > As you can see there is a failing test, this happens only when running together the tests of the suite start-typescript.test.js with error: child test left in queue: t.test should start the server with watch option in nested dir
> > No idea why
> 
> You need to reduce the test concurrancy to one, see https://node-tap.org/docs/api/parallel-tests/

I thought this was already set in `.taprc` with jobs=1
I also tried setting {buffered: false} in tests options and `process.env.TAP_BUFFERED=0`
And launching it from CLI setting -j1 and --jobs=1

Am I missing something?

fox1t:
The best we can do for the moment is to revert and use this PR in conjunction with https://github.com/danez/ts-example-app to itarate! :)
@alemagio Can you check this PR against that repo?

alemagio:
@fox1t 
Yeah I'm using that repo right now as reference :+1: 
At the moment it looks fine.
I'm still working to fix the tests :sweat_smile:

lib/watch/tsc-watcher.js
```diff
@@ -101,4 +104,28 @@ function reportWatchStatusChanged (diagnostic) {
   console.info(ts.formatDiagnostic(diagnostic, formatHost))
 }
 
+function requireTsConfig (configPath) {
+  const isModule = basename(configPath) === configPath
+  let configFile = require(isModule ? configPath : resolve(process.cwd(), configPath))
```
mcollina, line 109: I think the tsconfig can be a json5 with comments, right? So this won't work.

lib/watch/tsc-watcher.js
```diff
@@ -56,9 +56,12 @@ function watchMain (args, opts) {
     }
     origPostProgramCreate(program)
 
+    // We need to set the NODE_PATH variable in case node_modules are in a child directory
+    const env = Object.assign({}, process.env, { NODE_PATH: resolve(process.cwd(), 'node_modules') })
```
mcollina, line 60: This scares me quite a lot. Why is it needed?

alemagio: It scares me too, but the problem is that we can have a parent directory with `node_modules` and a child directory having `node_modules` too.
If the `outDir` is called lib (for example) and it is built in the parent dir we have:
- parent
  -- child
    --- node_modules
    --- src
  -- node_modules
  -- lib 
    --- index.js
If in index.js we require something which is in child/node_modules it is not found.

I believe a more soft approach might be to append the directory to the `NODE_PATH` instead of replacing it. What do you think?

lib/watch/tsc-watcher.js
alemagio: Ok, I'll add `require('json5/lib/register')`

lib/watch/tsc-watcher.js
mcollina: I would not touch `NODE_PATH` in any way.

climba03003:
Can we just spawn a child_process with ```ts-node```?
It should be the most simple solution and we dont need to deal with the staff like ```tsconfig```.
As the issue state, it work normally when he is using the ```ts-node``` command.

alemagio:
> Can we just spawn a child_process with `ts-node`?
> It should be the most simple solution and we dont need to deal with the staff like `tsconfig`.
> As the issue state, it work normally when he is using the `ts-node` command.
> 
> Simple solution for watch feature is to spawn the `tsc --watch` process then we start it normally using the original function as it compiled to `js`.

Using `ts-node` works, I just tried a different approach using the tsc API.

Anyway it looks like I'm stuck at the moment because of what explained [here](https://github.com/fastify/fastify-cli/pull/282#discussion_r507809890) 
I'm still working on this to find a solution.

If you guys want we can close this PR and simply proceed with the `ts-node` version or even leave the TS handling to the user.

climba03003:
> 
> 
> > Can we just spawn a child_process with `ts-node`?
> > It should be the most simple solution and we dont need to deal with the staff like `tsconfig`.
> > As the issue state, it work normally when he is using the `ts-node` command.
> > Simple solution for watch feature is to spawn the `tsc --watch` process then we start it normally using the original function as it compiled to `js`.
> 
> Using `ts-node` works, I just tried a different approach using the tsc API.
> 
> Anyway it looks like I'm stuck at the moment because of what explained [here](https://github.com/fastify/fastify-cli/pull/282#discussion_r507809890)
> I'm still working on this to find a solution.
> 
> If you guys want we can close this PR and simply proceed with the `ts-node` version or even leave the TS handling to the user.

I do not see any reason it will create that strange file structure. 
Even if we specify the ```outDir``` TypeScript will not compile the ```node_modules``` folder as we always excluded it.
If it is true that the file structure is created, the code can not be executed in first place even without TypeScript.

Issue #267, opened 2020-09-06
Support ESM modules


Eomm:
## 🚀 Feature Proposal

Support ESM module

## Motivation

In the node14 the ESM modules will be no longer experimental

## Example

```
// demo.mjs

import fs from 'fs'

export default function app (fastify, opts, next) {
  fastify.get('/', function (request, reply) {
    reply.send(fs.readFileSync('./package.json'))
  })

  next()
}
```
And then run it with `fastify start demo.mjs`

This feature should:

- load the arg file as ESM when the file ends with `mjs` or the `type` in the package.json is `module`

The watching feature should be updated accordingly.

To support node10, it will be needed a `--esm` `--no-esm` flag because it doesn't support the `type:module` loading as node>=12 does

mcollina:
For whoever takes this, https://github.com/fastify/fastify-autoload has all the necessary examples on how to load _both_ cjs and mjs.

danielglazer:
Hi, can I grab this? looking for my first contribution

mcollina:
go for it!

fox1t:
I would be happy to review that PR :)

seriousme:
Just wondering: would it not be easier to just always  `import` as of Node 14.
`import` will happily import .js , .cjs and .mjs

E.g, in [util.js](https://github.com/fastify/fastify-cli/blob/master/util.js#L40) something like:
```javascript
const hasESM = process.version.split(/[v|\./]/)[1] >= 14
const serverPlugin = hasESM ? await import(resolvedModulePath) : require(resolvedModulePath)
```

This way you don't need all the extra logic present in  https://github.com/fastify/fastify-autoload 
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import 'import()'  does not need the URL format so that would alleviate the need of importing the URL module.

Did I miss something ?
I plan to add a similar trick to fastify-openapi-glue, so any knowledge gained here would help me as well ;-)

Kind regards,
Hans

seriousme:
It turns out that `import` returns an object containing the export so to get it to work it needs to be something like:
```javascript
const hasESM = process.version.split(/[v|\./]/)[1] >= 14
const serverPlugin = hasESM ? (await import(resolvedModulePath)).default : require(resolvedModulePath)
```
Kind regards,
Hans

mcollina:
It likely would!

fox1t:
[...]

PR #288, opened 2020-10-30
support ESM modules

seriousme:
#### Checklist

- [X] documentation is changed or added

This PR adds support for ESM modules as discussed in #267 

Kind regards,
Hans


util.js
```diff
@@ -1,8 +1,11 @@
 const fs = require('fs')
 const path = require('path')
+const url = require('url')
 
 const resolveFrom = require('resolve-from')
 
+const hasESM = process.versions.node.split('.')[0] >= 14
```
mcollina, line 7: This is not true. It would work on Node 12. I think this detection should be more granular: see how it is done in fastify-autoload.

util.js
```diff
@@ -30,14 +33,14 @@ function isInvalidAsyncPlugin (plugin) {
...
-function requireServerPluginFromPath (modulePath) {
+async function requireServerPluginFromPath (modulePath) {
   const resolvedModulePath = path.resolve(process.cwd(), modulePath)
 
   if (!fs.existsSync(resolvedModulePath)) {
     throw new Error(`${resolvedModulePath} doesn't exist within ${process.cwd()}`)
   }
 
-  const serverPlugin = require(resolvedModulePath)
+  const serverPlugin = hasESM ? (await import(url.pathToFileURL(resolvedModulePath).href)).default : require(resolvedModulePath)
```
mcollina, line 43: I would prefer not to load cjs via esm.

README.md
```diff
@@ -109,6 +109,16 @@ module.exports = function (fastify, options, next) {
...
 ```
 
+Modules in EcmaScript Module format can be used on Node.js >= 14.0.0
+```js
+// plugin.mjs
+const plugin = function (fastify, options, next) {
+  fastify.get('/', (req, reply) => reply.send(options))
+  next()
+}
+export { plugin as default }
```
mcollina, line 119: I prefer `export default plugin`

util.js
seriousme: It is true that it works on >= 14. 
But it *also* works on >12.17 but not on 13 ;-)

I have seen how its done in fastify-autoload as you mentioned that already near the start of #267 

I tried to avoid the complexity both on code and in the docs by just starting at 14 as:
- 14 was the first version to support from ESM from day 1 without feature flags
- 14 is active LTS and 12 goes into LTS maintenance within a month
- ESM is still experimental anyway

And from your [comment](https://github.com/fastify/fastify-cli/issues/267#issuecomment-716348435) I got the impression that you agreed. 

Do you still want the additional complexity ?

util.js
seriousme: I tried the avoid the complexity present in fastify-autoload of having to redo nodes internal algorithm to determine what is CJS and ESM. If in a years time the node community changes the internal algorithm we need to change this code as well. That is what I was trying to avoid.

The [NodeJS documentation](https://nodejs.org/api/esm.html#esm_import_expressions) shows that 'import()' is allowed to be used to import CJS.
```js
import * as m from 'cjs';
console.log(m);
console.log(m === await import('cjs'));
// Prints:
//   [Module] { default: <module.exports> }
//   true
```
Do you still want the the additional complexity ?

README.md
seriousme: Fine, this was what VScode created on conversion to Module format.

seriousme: I have pushed this in c3e1ab7

seriousme: and fixed 'examples/ts-plugin-with-custom-options.mjs' in 324d47a

README.md
```diff
@@ -109,6 +109,16 @@ module.exports = function (fastify, options, next) {
 $ fastify start plugin.js -- --one
 ```
 
+Modules in EcmaScript Module format can be used on Node.js >= 14.0.0
```
mcollina, line 112: I think also v12.something right now.

examples/ts-plugin-with-custom-options.mjs
```diff
@@ -0,0 +1,8 @@
+// Module code is always strict mode code.
+// http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code
+
+const plugin = function (fastify, options, next) {
```
mcollina, line 4: I would make this an async named function instead

util.js
mcollina: ESM is going out of experimental in v12 in the upcoming release and it's already out of experimental in v14. We need the added complexity unfortunately :(

mcollina: You can also check if there is `global.import()` available.

util.js
mcollina: Yes. cjs over esm go through some other code branch.

mcollina (changes requested):

util.js
seriousme: I like the global.import() check but the only way that I know is:
```js
let hasESM = false
import('').catch(error => {
    if (error !== 'Not supported') {
        hasESM = true
    }
})
```
as `typeof import === 'function'` returns a syntax error :-(

The downside of the .catch trick is that it sends a warning to stderr on Node 12.18.3:
`(node:20501) ExperimentalWarning: The ESM module loader is experimental.`
Do you have a better way ?

examples/ts-plugin-with-custom-options.mjs
seriousme: Fixed both in the example and the README

README.md
seriousme: added >=12.17.0 <13.0.0

util.js
seriousme: added >=12.17.0 <13.0.0

util.js
```diff
@@ -30,14 +33,24 @@ function isInvalidAsyncPlugin (plugin) {
...
   const resolvedModulePath = path.resolve(process.cwd(), modulePath)
 
   if (!fs.existsSync(resolvedModulePath)) {
     throw new Error(`${resolvedModulePath} doesn't exist within ${process.cwd()}`)
   }
 
-  const serverPlugin = require(resolvedModulePath)
+  const modulePattern = /\.mjs$/i
+  let serverPlugin
+  if (modulePattern.test(modulePath)) {
```
mcollina, line 45: You also need to load `package.json` and check if `type: 'module'` is there. In that case the `.js` file extension will be used to load modules, and `.cjs` is for esm.

Read more at https://nodejs.org/api/packages.html#packages_package_json_and_file_extensions

mcollina (commented):
Good work! I just noticed we are missing an important case.

util.js
seriousme: I added the use case as requested including a test case.

One more question though:

I think its quite inconvenient that each and every project that uses plugins would need to added this additional complexity. 
Hence a single module that would abstract this away would be quite convenient imho, e.g. "importAny".
Where `importAny(file)` would conditionally use `import(file)` or `require(file)` depending on the rules depicted in https://nodejs.org/api/packages.html#packages_determining_module_system.

But when I look at "https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_interoperability_with_commonjs" it states:
```quote
An import statement can reference an ES module or a CommonJS module. Other file types such as JSON or native modules are not supported. For those, use module.createRequire().
```
So that would mean that `import()` already does what I would want `importAny` to do.

Although maybe not relevant anymore for fastify-cli, I am still curious to why you don't want to use `import` for CJS even when the docs recommend it for interoperability.

seriousme: Just noticed that I missed the alternative case: package.json says "module" but file says ".cjs"

fox1t:
I am going to review this in a couple of hours hopefully. Support for ESM could be really tricky these days :(

util.js
seriousme: just added the missing use case.

seriousme: It does prove that its a complex thing to do this right.

mcollina (approved):
lgtm

fox1t (approved):
LGTM!!!

Seems that we are not missing anything here. Still, fingers crossed since this could lead to awful bugs! :(

PR #300, opened 2020-12-17
feat: reload env when watch restarts the application

simoneb:
Reload the environment when the app restarts in watch mode. Haven't added any tests yet as there are none around watch mode. I would like to validate the idea before spending time testing this, which is not straightforward.

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         node: [ '10', '12', '14' ]
-    name: Node ${{ matrix.node }}
+    name: Node ${{ matrix.node }} on ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
     - name: Setup Node
--- a/args.js
+++ b/args.js
@@ -32,7 +32,7 @@ module.exports = function parseArgs (args) {
       watch: false,
       debug: false,
       debugPort: 9320,
-      'ignore-watch': 'node_modules build dist .git bower_components logs .swp',
+      'ignore-watch': 'node_modules build dist .git bower_components logs .swp .nyc_output',
       options: false,
       'plugin-timeout': 10 * 1000, // everything should load in 10 seconds
       lang: 'js'
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -4,34 +4,32 @@ const path = require('path')
 const cp = require('child_process')
 const chalk = require('chalk')
 const { arrayToRegExp } = require('./utils')
-
-const {
-  GRACEFUL_SHUT
-} = require('./constants.js')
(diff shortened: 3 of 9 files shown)


mcollina (changes requested):
Can you please add a unit test?

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         node: [ '10', '12', '14' ]
-    name: Node ${{ matrix.node }}
+    name: Node ${{ matrix.node }} on ${{ matrix.os }}
```
simoneb, line 12: so you know which os CI is running on, besides Node version

args.js
```diff
@@ -32,7 +32,7 @@ module.exports = function parseArgs (args) {
       watch: false,
       debug: false,
       debugPort: 9320,
-      'ignore-watch': 'node_modules build dist .git bower_components logs .swp',
+      'ignore-watch': 'node_modules build dist .git bower_components logs .swp .nyc_output',
```
simoneb, line 35: watch tests hang without this because tap is watching this folder too without the ignore

lib/watch/index.js
```diff
@@ -4,34 +4,32 @@ const path = require('path')
 const cp = require('child_process')
 const chalk = require('chalk')
 const { arrayToRegExp } = require('./utils')
-
```
simoneb, line 7: the changes in this file are necessary because otherwise `watch` uses globals which prevent multiple tests executing watch from running. I moved everything to local within the watch function

test/start.test.js
```diff
@@ -548,14 +549,56 @@ test('should start the server with watch options that the child process restart
...
-  await writeFile(tmpjs, 'hello fastify', { flag: 'a+' }) // chokidar watch can't caught change event in travis CI, but local test is all ok. you can remove annotation in local environment.
+  await writeFile(tmpjs, 'hello fastify', { flag: 'a+' }) // chokidar watch can't catch change event in CI, but local test is all ok. you can remove annotation in local environment.
   t.pass('change tmpjs')
 
   // this might happen more than once but does not matter in this context
   await once(fastifyEmitter, 'restart')
   t.pass('should receive restart event')
 })
 
+test('should reload the env on restart when watching', async (t) => {
```
simoneb, line 560: this test is similar to the existing one but I believe more robust, because it uses tap's native testdir to create a directory and change the cwd to that, so we're sure that the test is isolated from others. I know it introduces a different pattern from other tests, but I think it's better written in this way.

simoneb (commented):
@mcollina test added, see inline comments. most changes are to make test pass. the only real change needed to reload the .env is to move dotenv.config() to the function that starts fastify instead of having among the top level require statements

mcollina (approved):
lgtm

PR #304, opened 2020-12-22
fix: fix start bug in typescript template

### LINKED ISSUE #303: Typescript template does not work with start command
## 🐛 Bug Report

Simply running `npm run dev` of `npm run start` in a clean typescript template does not work:

image

Changing the import to `import { join } from "path"` fixes the issue but another one comes out:

image

## Your Environment

- *node version*: 12 and 14
- *fastify version*: >= 3
- *os*: Linux

Adding `esModuleInterop` to tsconfig.json fixes both.

alemagio:
#### Checklist

Fixes #303

### DIFF
--- a/templates/app-ts/tsconfig.json
+++ b/templates/app-ts/tsconfig.json
@@ -1,7 +1,8 @@
 {
     "extends": "fastify-tsconfig",
     "compilerOptions": {
-        "outDir": "dist"
+        "outDir": "dist",
+        "esModuleInterop": true
     },
     "include": ["src/**/*.ts"]
 }


kibertoad:
This feels like a wrong solution, TBH. Isn't there a different way to address this? Bumping Node types dep version doesn't help? Using named import doesn't help?

kibertoad (changes requested):
See comments

mcollina:
Also.. add a test :).

alemagio:
> This feels like a wrong solution, TBH. Isn't there a different way to address this? Bumping Node types dep version doesn't help? Using named import doesn't help?

I believe it might be an issue somehow related to `fastify-autoload`, I tried making a simple app.ts file and registering manually the plugin. That works.

I also tried changing the plugin according to the latest [doc](https://www.fastify.io/docs/latest/TypeScript/#plugins) but it does not seem to work.

PS
Test incoming too.

kibertoad:
`I believe it might be an issue somehow related to fastify-autoload`

I doubt it. `fastify-autoload` exports look like this:
```
declare const fastifyAutoload: FastifyPlugin<AutoloadPluginOptions>
export default fastifyAutoload
```

This should work fine with esModuleInterop off.
As I've said, based on the error that you've mentioned in original issue, `import path from 'path';`  is the culprit. Can you try changing it to `import { join } from 'path';`?

alemagio:
@kibertoad 
 Second screenshot in issue shows what happens changing the import of `path`.
 @types/node is at 14.14.14 which is the latest.

By saying fastify-autoload might be the issue I meant the way it requires plugins, not fastify-autoload itself. But, again, not sure about that.

mcollina:
If it's an issue with fastify-autoload, then we should fix it in fastify-autoload.

alemagio:
Changing the app.ts file to this works:

```
import {AutoloadPluginOptions} from 'fastify-autoload';
import { FastifyPluginAsync } from 'fastify';
import support from './plugins/support';

export type AppOptions = {
  // Place your custom options for app below here.
} & Partial<AutoloadPluginOptions>;

const app: FastifyPluginAsync<AppOptions> = async (
    fastify,
    opts
): Promise<void> => {
  void fastify.register(support)
};

export default app;
```

@mcollina 
I am not sure actually, at first I thought the issue was in the template of the CLI.

mcollina:
Any updates on this?

alemagio:
Sorry, holidays delay :sweat_smile: 
#306 fixed compilation error but still getting the same error when running `npm run dev`.

I also found another error when running `npm run test`
```
test/helper.ts:5:8 - error TS1259: Module '"/home/alemagio/projects/fastify-cli/cane/node_modules/@types/tap/index"' can only be default-imported using the 'esModuleInterop' flag

5 import tap from 'tap';
         ~~~

  node_modules/@types/tap/index.d.ts:13:1
    13 export = tap;
       ~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

```
Which looks related to the fact that we are importing `tap` using default import in helper.ts

kibertoad:
@alemagio Thanks for the update! Will finally have time to investigate further tomorrow.

mcollina:
Ping, any updates?

mcollina (approved):
lgtm, we should land this

alemagio:
@mcollina 
This was fixed by [this](https://github.com/fastify/fastify-autoload/pull/126).
It can be closed without merging I believe.

alemagio:
@mcollina 
This was fixed by [this](https://github.com/fastify/fastify-autoload/pull/126).

fox1t:
How https://github.com/fastify/fastify-autoload/pull/126 fixed this? I am missing some pieces here.
To recap: if a package exports a namespace (ex. `export = path`), we should import it like `import * as path from 'path` if `esModuleInterop: false`. One important thing to mention is that if we use `esModuleInterop: true`, we force our users to set it to true too. This is not a big deal since `esModuleInterop: true` is the recommended one nowadays.

alemagio:
> How [fastify/fastify-autoload#126](https://github.com/fastify/fastify-autoload/pull/126) fixed this? I am missing some pieces here.
> To recap: if a package exports a namespace (ex. `export = path`), we should import it like `import * as path from 'path` if `esModuleInterop: false`. One important thing to mention is that if we use `esModuleInterop: true`, we force our users to set it to true too. This is not a big deal since `esModuleInterop: true` is the recommended one nowadays.

My bad, I thought we were still talking about the issue this PR was related to.

PR #305, opened 2020-12-22
Fix TS compilation error and add compilation to testing step

kibertoad:
ref #303

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint:fix": "standard --fix",
     "unit": "tap test/*.test.js templates/app/test/*/*.test.js templates/app-ts/test/*/*.test.ts templates/plugin/test/*.test.js",
     "test": "npm run lint && npm run unit && npm run test:typescript",
-    "test:typescript": "tsd templates/plugin"
+    "test:typescript": "tsd templates/plugin && tsc --project templates/app-ts/tsconfig.json"
   },
   "keywords": [
     "fastify",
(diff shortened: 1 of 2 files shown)


kibertoad:
@alemagio @mcollina This resolves the problem of project not compiling. Still need to add test for app failing on startup, but I believe that to be a separate issue.

alemagio:
@kibertoad 
I created a [repo](https://github.com/alemagio/fastify-cli-ts-issue) with the same (hopefully) conditions I have.

kibertoad:
superseded by #306

kibertoad:
@alemagio Check #306, I tried to reproduce the issue in a test, but app starts just fine.
What do I do in `https://github.com/alemagio/fastify-cli-ts-issue`? just `npm run start`?

alemagio:
> @alemagio Check #306, I tried to reproduce the issue in a test, but app starts just fine.
> What do I do in `https://github.com/alemagio/fastify-cli-ts-issue`? just `npm run start`?

I have issues with both `npm run start` and `npm run dev`, same problem.

kibertoad:
@alemagio Yeah, failing locally for me as well. Considering that direction invocation of `app.ts` works, I suspect problem is with start.js being incompatible with app.ts somehow. I'll look deeper into it.

alemagio:
Using the very same command but without fastify autoload works as shown [here](https://github.com/fastify/fastify-cli/pull/304#issuecomment-749608855) so I don't think the issue is in start.js.
Don't you think?

kibertoad:
@alemagio But using exact same code that calls auto-load from `test/templates/app-ts.test.ts`, bypassing start.js, works, so situation is kinda muddy :). So it seems like a combination of the two, and problem could be on either, or both, sides.

PR #306, opened 2020-12-22
Fix TS compilation error.

### LINKED ISSUE #303: Typescript template does not work with start command

kibertoad:
Also includes tests both for compilation and for trying to run the app from app-ts.

refs #303

#### Checklist

- [] documentation is changed or added

### DIFF
--- a/templates/app-ts/src/app.ts
+++ b/templates/app-ts/src/app.ts
@@ -1,4 +1,4 @@
-import path from 'path';
+import { join } from 'path';
 import AutoLoad, {AutoloadPluginOptions} from 'fastify-autoload';
 import { FastifyPluginAsync } from 'fastify';
 
@@ -18,17 +18,18 @@ const app: FastifyPluginAsync<AppOptions> = async (
   // those should be support plugins that are reused
   // through your application
   void fastify.register(AutoLoad, {
-    dir: path.join(__dirname, 'plugins'),
+    dir: join(__dirname, 'plugins'),
     options: opts
   })
 
   // This loads all plugins defined in routes
   // define your routes in one of these
   void fastify.register(AutoLoad, {
-    dir: path.join(__dirname, 'routes'),
+    dir: join(__dirname, 'routes'),
     options: opts
   })
 
 };
 
 export default app;
+export { app }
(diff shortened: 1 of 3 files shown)


templates/app-ts/src/app.ts
```diff
@@ -18,17 +18,18 @@ const app: FastifyPluginAsync<AppOptions> = async (
...
   void fastify.register(AutoLoad, {
-    dir: path.join(__dirname, 'routes'),
+    dir: join(__dirname, 'routes'),
     options: opts
   })
 
 };
 
 export default app;
+export { app }
```
mcollina, line 35: why this change?

kibertoad: Fair point. This is my pet peeve with the lack of named exports to go with the default, so for convenience of writing a test I added it, but if you prefer having just a default, I can remove it.

mcollina: Nice ok. Maybe in the test you can also check the default export?

kibertoad: Done.

test/templates/app-ts.test.ts
```diff
@@ -0,0 +1,55 @@
+import { fastify } from 'fastify'
+import { test } from 'tap'
+const sgetOriginal = require('simple-get').concat
+
+import appDefault, { app } from '../../templates/app-ts/src/app'
+
+const PORT = 3001;
```
mcollina, line 7:
```suggestion
const PORT = 0;
```

mcollina: use a random a picked port

kibertoad: Done. Needed slight adjustment in test itself as well.

mcollina (approved):
lgtm

Issue #314, opened 2021-02-13
Load a file before everything else


delvedor:
## 🚀 Feature Proposal

Offer a way to run a specific file before everything else, à la `node -r foo index.js`

## Motivation

Most APM vendors needs to be posed before anything else, which is currently not possible with the CLI.

## Example

```
fastify start -r apm.js app.js
```

mcollina:
sure!

anthonyringoet:
Merged to master so can be closed afaik.

mhamann:
@mcollina any chance we can get a release with this in it?

PR #316, opened 2021-02-18
Add option to load module before anything else #314

anthonyringoet:
As described in #314.

Some notes:
- `-r` option was already in use for `prefix` so took `b`/`before-module` (like existing `logging-module` option). Feel free to request other naming if preferred.
- Requiring the actual module seemed exactly the same as the existing `logging-module`. In an effort to create a function out of it, I abstracted the core part of `util.js's` `requireServerPluginFromPath` to `requireModule`. This leads to some minor differences in textual error output. Behavior of when errors are thrown is still the same though.

#### Checklist

### DIFF
--- a/start.js
+++ b/start.js
@@ -3,7 +3,6 @@
 'use strict'
 
 const assert = require('assert')
-const path = require('path')
 const split = require('split2')
 const PinoColada = require('pino-colada')
 const pump = require('pump')
@@ -15,7 +14,8 @@ const {
   exit,
   requireFastifyForModule,
   requireServerPluginFromPath,
-  showHelpForCommand
+  showHelpForCommand,
+  requireModule
 } = require('./util')
 
 let Fastify = null
@@ -59,8 +59,16 @@ function stop (message) {
 
 async function runFastify (args) {
   require('dotenv').config()
-
   const opts = parseArgs(args)
+
+  if (opts.beforeModule) {
+    try {
+      await requireModule(opts.beforeModule)
+    } catch (e) {
+      return module.exports.stop(e)
+    }
+  }
+
   opts.port = opts.port || process.env.PORT || 3000
 
   loadModules(opts)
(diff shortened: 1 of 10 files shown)


mcollina:
I think we should keep the same option name of Node.js, following the principle of least surprise (we will bump the major).

start.js
```diff
@@ -59,8 +59,16 @@ function stop (message) {
 
 async function runFastify (args) {
   require('dotenv').config()
-
   const opts = parseArgs(args)
+
+  if (opts.beforeModule) {
```
mcollina, line 64: how would this be handled for the `watch` option?

mcollina (changes requested):
This _might_ not be exactly what @delvedor needs. If I understand correctly, he needs this to load Elastic APM before everything else. 

Specifically, this should be loading only commonjs and not ESM, and possibly done earlier in the process startup (right after argument parsing.

anthonyringoet:
Makes sense and will make it a lot more straightforward if it can move a level up towards the `start` function. Will look into these changes.

start.js
anthonyringoet: Watch option checks for changes in `process.cwd()` so currently for both this addition and the custom logger module it won't be triggered if it's outside of the current working directory. Should both of these files be watched explicitly as well?

mcollina (approved):
lgtm

climba03003 (commented):
I think the test case about `require` and `fail to require` is missing. 
1. `require` is really done before the `fastify` is running
2. the behavior when `require is fail to load`

The first test is important as it is the actual function is this PR.

zekth (changes requested):
Like @climba03003 said, missing the tests cases of require working and not working.

Otherwise lgtm

anthonyringoet:
Added support for multiple `-r` flags (same as node cli) and added tests.

@climba03003 How would you go about testing if the require(s) are really done before the fastify server is running?

climba03003:
> 
> 
> Added support for multiple `-r` args (same as node cli) and added tests.
> 
> @climba03003 How would you go about testing if the require(s) are really done before the fastify server is running?

Just like your other test case, you can mutate global variables. Then, you can execute `tap` inside the server files.
It can ensure you receive a the things you want in the server.

I think we can expect, if `required files` use `top-level async` method. The order will break? should this be tested or mention?

anthonyringoet:
@climba03003 Like the previous commit?

The top-level async behavior is the same with `node -r ./some-async-file.js` afaik.

climba03003 (approved):
Good to know that it is the same as `node -r`.

mcollina:
@zekth is this good to land now?

zekth (approved):
good work!

Issue #317, opened 2021-02-18
Add graceful shutdown


mcollina:
## 🚀 Feature Proposal

Most users would expect to have a graceful shutdown sequence, where all the current requests are completed before exiting the process.

[close-with-grace](http://npm.im/close-with-grace) implement such functionality, so I propose we use it inside `fastify start` and inside the server generated by `fastify eject`.

StarpTech:
It would be great to make it part of [`fastify.close`](https://github.com/fastify/fastify/blob/master/docs/Server.md#close) so plugin maintainers can ensure that their resources are cleaned gracefully.

mcollina:
I think the idea would be:

```js
const closeWithGrace = require('close-with-grace')

closeWithGrace({ delay: 500 }, async function ({ signal, err, manual }) {
  if (err) {
    fastify.log.error(err)
  }
  await fastify.close()
})
```

StarpTech:
That's great but every new project has to respect that. The plugin so soo small but would have a huge impact.

mcollina:
> That's great but every new project has to respect that. 

I know. Unfortunately adding global state would make it harder to reuse and test. As an example, `tap` adds `uncaughtException` and `unhandledRejection` handlers to catch errors that happens inside the tests. Other test frameworks, libraries and other modules adds those.

StarpTech:
Good points. I agree that can be tricky.

inigo-garcia:
I started to integrate this feature and after running some tests, I think we may need to remove the listeners once the server gets closed. The library implements a method called `closeWithGrace.uninstall()`. What do you think?

PR #318, opened 2021-02-18
feat: added graceful shutdown (#317) 

inigo-garcia:
Issue link: https://github.com/fastify/fastify-cli/issues/317

#### Checklist

### DIFF
--- a/start.js
+++ b/start.js
@@ -8,6 +8,7 @@ const split = require('split2')
 const PinoColada = require('pino-colada')
 const pump = require('pump')
 const isDocker = require('is-docker')
+const closeWithGrace = require('close-with-grace')
 const listenAddressDocker = '0.0.0.0'
 const watch = require('./lib/watch')
 const parseArgs = require('./args')
@@ -125,6 +126,18 @@ async function runFastify (args) {
   }
 
   await fastify.register(file.default || file, opts.pluginOptions)
+
+  const closeListeners = closeWithGrace({ delay: 500 }, async function ({ signal, err, manual }) {
+    if (err) {
+      fastify.log.error(err)
+    }
+    await fastify.close()
+  })
+
+  await fastify.addHook('onClose', (instance, done) => {
+    closeListeners.uninstall()
+    done()
+  })
 
   if (opts.address) {
     await fastify.listen(opts.port, opts.address)
--- a/templates/eject-ts/server.ts
+++ b/templates/eject-ts/server.ts
@@ -7,6 +7,9 @@ dotenv.config();
 // Require the framework
 import Fastify from "fastify";
 
+// Require library to exit fastify process, gracefully (if possible)
+import closeWithGrace from "close-with-grace";
+
 // Instantiate Fastify with some config
 const app = Fastify({
   logger: true,
@@ -14,6 +17,19 @@ const app = Fastify({
 
 // Register your application as a normal plugin.
 app.register(import("./app"));
+
+// delay is the number of milliseconds for the graceful close to finish 
+const closeListeners = closeWithGrace({ delay: 500 }, async function ({ signal, err, manual }) {
+  if (err) {
+    app.log.error(err)
+  }
+  await app.close()
+})
+
+app.addHook('onClose', async (instance, done) => {
+  closeListeners.uninstall()
+  done()
+})
 
 // Start listening.
 app.listen(process.env.PORT || 3000, (err: any) => {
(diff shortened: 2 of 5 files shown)


test/start.test.js
```diff
@@ -9,6 +9,7 @@ const semver = require('semver')
 const baseFilename = path.join(__dirname, 'fixtures', `test_${crypto.randomBytes(16).toString('hex')}`)
 const { fork } = require('child_process')
 const moduleSupport = semver.satisfies(process.version, '>= 14 || >= 12.17.0 < 13.0.0')
+require('events').EventEmitter.defaultMaxListeners = 0
```
kibertoad, line 12: can you add comment to this line because it's not self-explanatory why it's there?

inigo-garcia: I was running into these warnings when I was testing the implementation. `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGTERM listeners added to [process]. Use emitter.setMaxListeners() to increase limit`

I tried with: afterEachTest

kibertoad: If we are actually leaking listeners somewhere, that's rather bad, we are likely to cause warnings for user tests downstream then.

templates/eject-ts/server.ts
```diff
@@ -7,6 +7,9 @@ dotenv.config();
 // Require the framework
 import Fastify from "fastify";
 
+// Require libraty to exit fastify process, gracefully (if possible)
```
simoneb, line 10:
```suggestion
// Require library to exit fastify process, gracefully (if possible)
```

templates/eject/server.js
```diff
@@ -6,6 +6,9 @@ require('dotenv').config()
 // Require the framework
 const Fastify = require('fastify')
 
+// Require libraty to exit fastify process, gracefully (if possible)
```
simoneb, line 9:
```suggestion
// Require library to exit fastify process, gracefully (if possible)
```

simoneb (approved):
code wise lgtm, does it work? :)

templates/eject/server.js
inigo-garcia: Thanks! I didn't notice it

inigo-garcia:
i tried sending SIGTERM and SIGINT signals and it handles them correctly.

simoneb:
@inigo-garcia please link the issue in the PR description

mcollina (changes requested):
This needs tests

test/start.test.js
mcollina: Unfortunately this means that we are not cleaning things up properly. You might need to call `uninstall` - see https://github.com/mcollina/close-with-grace#return-values.

mcollina (changes requested):

test/start.test.js
inigo-garcia: Thanks for the feedback, I've included the `uninstall` method inside of `onClose` hook and we don't get any memory leak warnings

start.js
```diff
@@ -126,6 +127,18 @@ async function runFastify (args) {
...
+  const closeListeners = closeWithGrace({ delay: 500 }, async function ({ signal, err, manual }) {
+    if (err) {
+      fastify.log.error(err)
+    }
+    await fastify.close()
+  })
+
+  await fastify.addHook('onClose', async (instance, done) => {
+    closeListeners.uninstall()
+    done()
```
mcollina, line 139: Don't call `done()` when using `async`.

mcollina (changes requested):

start.js
inigo-garcia: Thanks @mcollina , I'll update it

mcollina (approved):
lgtm

inigo-garcia:
@mcollina  @kibertoad @simoneb Thanks for your feedback

kibertoad (approved):

templates/eject/server.js
```diff
@@ -15,6 +18,19 @@ const app = Fastify({
...
 
+// delay is the number of milliseconds for the graceful close to finish
+const closeListeners = closeWithGrace({ delay: 500 }, async function ({ signal, err, manual }) {
+  if (err) {
+    app.log.error(err)
+  }
+  await app.close()
+})
+
+app.addHook('onClose', async (instance, done) => {
```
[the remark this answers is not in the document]
simoneb: no special reason I suppose

PR #326, opened 2021-03-15
Change .mjs to .js and add note about "type": "module"

benatkin:
`fastify-cli` is really cool! I want to use esnext with no build step for my current project, and was glad to find that fastify-cli has no conflict with node's new ability to run es modules with a `.js` extension when `"type": "module"` is set in package.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,8 +121,9 @@ export default async function plugin (fastify, options) {
 }
 ```
 
-(You need to set `"type": "module"` in `package.json` for this to work with a `.js`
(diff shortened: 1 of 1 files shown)


mcollina (commented):
Yes, it's totally supported. Could you clarify that also `.mjs` is supported? Both works.

benatkin:
@mcollina I clarified it and added a link to the node.js documentation about it. Thanks!

mcollina (approved):
lgtm

Issue #327, opened 2021-03-16
Environment variable FASTIFY_WATCH is not honored


mariusa:
https://github.com/fastify/fastify-cli/issues/151 reappeared

mariusa:
https://github.com/fastify/fastify-cli/pull/329 is the correct fix

PR #328, opened 2021-03-16
keep dotenv at top in order for watch env variable to be parsed before start

mariusa:
https://github.com/fastify/fastify-cli/issues/327

#### Checklist

### DIFF
--- a/start.js
+++ b/start.js
@@ -2,6 +2,8 @@
 
 'use strict'
 
+require('dotenv').config() // keep at top in order for watch env variable to be parsed before start
+
 const assert = require('assert')
 const split = require('split2')
 const PinoColada = require('pino-colada')
@@ -59,7 +61,6 @@ function stop (message) {
 }
 
 async function runFastify (args) {
-  require('dotenv').config()
   const opts = parseArgs(args)
   if (opts.require) {
     if (typeof opts.require === 'string') {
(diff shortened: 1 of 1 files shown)


mcollina:
@simoneb you changed this in https://github.com/fastify/fastify-cli/pull/300. Wdyt of this change?

simoneb:
I'll need to have a better look because I don't understand why the test I added in that PR is not failing with this change which basically reverts that.

simoneb:
@mariusa can you please add a test so we avoid this regression once again? Unfortunately I think moving it back to the top breaks what was implemented in #300 , despite the test not catching it, so I don't think moving is good enough. We may actually need to have both, even though testing those cases is not straightforward.

mariusa:
Sorry Simone, I just got started with fastify, I won't be able to make a test. Also, I had to make twice this trivial PR. First time I used vscode with my auto-formatting settings, realized only when opening the PR. I had to edit it again with a plain text editor. 

> We may actually need to have both

I've added it back too, that makes sense.

simoneb:
See https://github.com/fastify/fastify-cli/pull/329. The change is not obvious in order to support what I introduced in #300 and to undo the regression. Longer explanation in there. I couldn't write an automated test because getting this right is not straightforward.

PR #329, opened 2021-03-16
fix: make both fastify and plugin env vars work

### LINKED ISSUE #327: Environment variable FASTIFY_WATCH is not honored

simoneb:
Fixes #327 

This change makes `fastify start` work for both scenarios:

1. using FASTIFY_* env vars from `.env` file, e.g. `FASTIFY_WATCH`
2. support reloading after `.env` file change in watch mode so that the app sees the updated env vars

### Why?

- the version before #300 was loading `.env` at `start.js` module initialization, thereby supporting only case 1) and not case 2) because even if the module is reloaded within the same process, dotenv doesn't overwrite env vars which are already set
- the version after #300 was loading `.env` too late, thereby supporting only case 2) but not case 1)

### What's missing

- an automated test, which is not straightforward to write

### How to test manually

- `cd examples`
- add `FASTIFY_WATCH=true` to `.env`
- `node ../cli.js start plugin-with-env.js` (note we're not providing `-w`)
- see that changing the value of `GREETING` in `.env` reloads the app (meaning that 1) is supported) and that the change in `GREETING` is reflected in the response of the app running at localhost:3000 (which shows that case 2) works)

Without the change in this PR, either of the 2 cases won't work.

#### Checklist

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -39,7 +39,7 @@ const watch = function (args, ignoreWatch, verboseWatch) {
 
   const run = (event) => {
     const childEvent = { childEvent: event }
-    const env = Object.assign({}, process.env, childEvent)
+    const env = Object.assign({}, process.env, childEvent, require('dotenv').config().parsed)
 
     const _child = cp.fork(forkPath, args, {
       env: env,
--- a/start.js
+++ b/start.js
@@ -2,6 +2,7 @@
 
 'use strict'
 
+require('dotenv').config()
 const assert = require('assert')
 const split = require('split2')
 const PinoColada = require('pino-colada')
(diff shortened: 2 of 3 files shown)


mcollina:
I'd go ahead and land this without tests as those watch tests are really tricky. Can you add some comments everywhere so we do not regress on this again?

simoneb:
I'd like to give this a try at least :)

simoneb:
This is now ready, I added a test to cover scenario 1) and modified an existing test which was supposed to cover case 2) but didn't

Eomm (approved):

mcollina (approved):
lgtm

PR #352, opened 2021-04-27
fix: watcher ignores ts files in typescript projects

alemagio:
#### Checklist

This patch was included as a comment [here](https://github.com/fastify/fastify-cli/issues/246#issuecomment-700819896), it should be included in the template and be like that out of the box.

Without this change we have a weird behavior between the typescript watcher and the CLI watcher.
CLI watcher basically restart before the TS compiler finishes its job so the changes are not visible immediately.


mcollina (approved):
lgtm

climba03003 (approved):
LGTM. it can ensure `fastify-cli` only restart when the compiled code change.

alemagio:
> LGTM. it can ensure `fastify-cli` only restart when the compiled code change.

Exactly

RafaelGSS (approved):
LGTM. Good catch!

PR #373, opened 2021-05-18
fix: skip watch test on github ci

simoneb:
Skip one test on GitHub CI. Same tests consistently succeeds locally. It seems other tests have had the same problem over time and are already skipped.

#### Checklist

### DIFF
--- a/test/start.test.js
+++ b/test/start.test.js
@@ -606,7 +606,7 @@ test('should start the server with watch and verbose-watch options that the chil
   t.ok(spy.calledOnce, 'should print a console message on file update')
 })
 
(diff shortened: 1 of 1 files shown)


mcollina:
I think this might be the only sane approach.

simoneb:
> I think this might be the only sane approach.

I tried a few options to no avail, there seems to be a race condition whereby the test directory dynamically created during tests is still locked when tap tries to clean it up.

mcollina (approved):
lgtm

simoneb:
let's retry the build when gh actions recover

Issue #426, opened 2021-11-10
How to debug typescript project (VScode)


ilijaNL:
Hello,

I wonder if it is possible to use the debugger (VSCode or Chrome) on a typescript project generated with fastify-cli.

Thanks

giacomorebonato:
It's possible to debug in VSCode by creating a `launch.json` file and picking the preset for **Node.js**.

![Screenshot 2021-11-11 at 11 25 51](https://user-images.githubusercontent.com/630796/141282283-931c49dd-9f89-4b9a-af76-0ec02c3b4ae3.png)

This is for example the `.vscode/launch.json` file that I get generated automatically:

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "pwa-node",
            "request": "launch",
            "name": "Launch Program",
            "skipFiles": [
                "<node_internals>/**"
            ],
            "program": "${workspaceFolder}/src/app.ts",
            "preLaunchTask": "tsc: build - tsconfig.json",
            "outFiles": [
                "${workspaceFolder}/dist/**/*.js"
            ]
        }
    ]
}
```

Sourcemaps also needs to be enabled in `tsconfig.json`

```json
{
    "extends": "fastify-tsconfig",
    "compilerOptions": {
        "outDir": "dist",
        "sourceMap": true
    },
    "include": ["src/**/*.ts"]
}
```

I tested this by putting a breakpoint on a project that I created with `fastify generate my-project --lang=ts`.

giacomorebonato:
@ilijaNL could you check if this implementation achieves what you're looking for?

https://github.com/fastify/fastify-cli/pull/427#pullrequestreview-803935839

ilijaNL:
> @ilijaNL could you check if this implementation achieves what you're looking for?
> 
> [#427 (review)](https://github.com/fastify/fastify-cli/pull/427#pullrequestreview-803935839)
Hmm cant get working on my project, npm run dev works fine but when starting with vs code it just closes after few seconds.

One thing about the suggestion is that it is not supporting watchers, and in my case always debug during development (using watch function)

giacomorebonato:
@ilijaNL do you have sample project where I can reproduce the issue?
It works for me when starting a new project from scratch.

[...]

PR #427, opened 2021-11-11
feat: config to debug TypeScript in vscode

giacomorebonato:
I am proposing to add these changes to both JavaScript and TypeScript templates to allow code debugging in VSCode out of the box.
Following the request on this issue: https://github.com/fastify/fastify-cli/issues/426

Screenshot 2021-11-11 at 11 59 45

## Steps for testing

- on this branch
- `node ./cli.js generate ../my-fastify-ts --lang ts` or `node ./cli.js generate ../my-fastify`
- `cd ../my-fastify-ts && npm install`

Now you should see the option "Launch Fastify" in the "Run and Debug" section of VSCode:

Screenshot 2021-11-11 at 16 56 15

#### Checklist

### DIFF
--- a/templates/app-ts/.vscode/launch.json
+++ b/templates/app-ts/.vscode/launch.json
@@ -8,7 +8,8 @@
             "skipFiles": [
                 "<node_internals>/**"
             ],
-            "program": "${workspaceFolder}/src/app.ts",
+            "program": "${workspaceRoot}/node_modules/.bin/fastify",
+            "args": ["start", "-l", "info", "./dist/app.js"],
             "preLaunchTask": "tsc: build - tsconfig.json",
             "outFiles": [
                 "${workspaceFolder}/dist/**/*.js"
--- a/templates/app-ts/.vscode/launch.json
+++ b/templates/app-ts/.vscode/launch.json
@@ -8,8 +8,8 @@
             "skipFiles": [
                 "<node_internals>/**"
             ],
-            "program": "${workspaceRoot}/node_modules/.bin/fastify",
-            "args": ["start", "-l", "info", "./dist/app.js"],
+            "program": "${workspaceRoot}/node_modules/.bin/nodemon",
+            "args": ["--exec", "npm", "run", "dev"],
             "preLaunchTask": "tsc: build - tsconfig.json",
             "outFiles": [
                 "${workspaceFolder}/dist/**/*.js"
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

simoneb (commented):
Shall we consider doing the same for the js version to keep consistency across both?

giacomorebonato:
Thanks for pointing out @simoneb .
I added the launch.json file also to the standard template.
A correction to the initial script was needed and I have now re-tested both.

simoneb:
> Thanks for pointing out @simoneb . I added the launch.json file also to the standard template. A correction to the initial script was needed and I have now re-tested both.

I don't see any changes

giacomorebonato:
Pushed now, sorry about that

templates/app-ts/.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+    "configurations": [
+        {
+            "type": "pwa-node",
+            "request": "launch",
+            "name": "Launch Fastify",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceRoot}/node_modules/.bin/fastify",
+            "args": ["start", "-l", "info", "./dist/app.js"],
```
simoneb, line 12: question: this may be a silly question. is it correct that the launch command runs on the dist folder, which I assume is the transpiled version of the app? Should it not execute on the original sources and invoke the same npm script?

giacomorebonato: This is the correct usage when sourcemaps are enabled, that are referencing the original ts files.
This allows placing breakpoints on ts files:

Screenshot 2021-11-11 at 16 22 52

I don't know inside out the implementation that allows this integration, but it works and it is a similar concept that applies also when debugging a frontend bundle with sourcemaps.

I know that if we add [ts-node](https://www.npmjs.com/package/ts-node) as a dependency we can achieve what you're pointing out.
I preferred to leave this dependency out because it starts to get opinionated, but it could be brought in if anyone has a strong point for going this way.

simoneb (approved):
LGTM, can you ask the original reporter of the issue to check that this would solve his problem?

giacomorebonato:
I added changes for having the debugger to automatically recomplile and restart if there's any change to a file.

This feature was missing compared to running the `npm run dev` command from terminal.

simoneb:
This should be good to go in

Eomm (approved):

Issue #416, opened 2021-10-09
Support build interface


Eomm:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Right now the test doc suggests to users to create a factory to build the fastify instance:

https://www.fastify.io/docs/latest/Testing/

```
function build(opts={}) {
  const app = fastify(opts)
  app.get('/', async function (request, reply) {
    return { hello: 'world' }
  })

  return app
}
```

This interface is easier to test for users and configure:

```
test('my test', t => {
  const fastify = build({})
  fastify.ready()
})
```

Instead, the `fastify-cli` interface requires  a simple plugin interface that force the user to write an helper that must know how `fastify-cli` works to instantiate it correctly.

```
function helper() {
   const app =Fastify(myApplication.config)
   app.register(myApplication)
   return app
}
```

### Motivation

So, this double interfaces does not fit togheter for newcomers and force the user to search for a pattern, instead of help them to find it.

For this reason, I think we should:

- expose the `build()` function that return a fastify intance when you `require('fastify-cli')`
- OR support the interface `build()` returned by the user, instead of the simple plugin interface

What do you think?

### Example

see above

mcollina:
I think we should expose some helpers for tests and other programmatic access.

PR #446, opened 2022-01-09
feat: add test helper

Eomm:
closes https://github.com/fastify/fastify-cli/issues/416

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,8 +325,9 @@ run `fastify docs` to get started.
 
 ## Test helpers
 
-When you use `fastify-cli` to run your project you need a way to test your application.
+When you use `fastify-cli` to run your project you need a way to load your application because you can run the CLI command.
 To do so, you can use the this module to load your application and give you the control to write your assertions.
+These utilities are async functions that you may use with the [`node-tap`](https://www.npmjs.com/package/tap) testing framework.
 
 There are two utilities provided:
 
@@ -335,7 +336,7 @@ There are two utilities provided:
 
 Both of these utilities have the `function(arg, pluginOptions)` parameters:
 
-- `arg`: is a string or a string array within the same arguments passed to the `fastify-cli` command.
+- `cliArgs`: is a string or a string array within the same arguments passed to the `fastify-cli` command.
 - `pluginOptions`: is an object containing the options provided to the started plugin (eg: `app.js`).
 
 ```js
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ const { build, listen } = require('fastify-cli/helper')
 const { test } = require('tap')
 test('test my application', async t => {
   const argv = ['app.js']
-  const app = await helper.build(argv, {
+  const app = await build(argv, {
     extraParam: 'foo'
   })
   t.teardown(() => app.close())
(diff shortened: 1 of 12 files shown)


mcollina (commented):
Can you integrate this in the generated scaffold?

Eomm:
@fastify/typescript 
I can't find a way to let this PR work.

```
templates/app-ts/test/routes/example.test.ts 2> TSError: ⨯ Unable to compile TypeScript:
templates/app-ts/test/routes/example.test.ts 2> templates/app-ts/test/helper.ts(3,23): error TS7016: Could not find a declaration file
for module 'fastify-cli/helper'. '/Users/mspigolon/workspace/fastify-cli/node_modules/fastify-cli/helper.js' implicitly has an 'any'
type.
```

Should I add the typings to this project?

Eomm:
@mcollina the PR is ready.

It does not update the TS template. I would open an issue for it, or I need some help from @fastify/typescript to make it works

templates/app/test/routes/root.test.js
```diff
@@ -1,10 +1,10 @@
 'use strict'
 
 const { test } = require('tap')
-const { build } = require('../helper')
+const { buildApplication } = require('../helper')
 
 test('default root route', async (t) => {
-  const app = build(t)
+  const app = await buildApplication(t)
```
mcollina, line 7: why do you need to await this? It's not strictly needed by fastify.

README.md
```diff
@@ -323,6 +323,41 @@ The documentation is rendered using an interactive terminal session that you can
 
 run `fastify docs` to get started.
 
+## Test helpers
+
```
mcollina, line 327: make it clear that the test helper are only for tap and not jest/mocha/tape etc.

mcollina (changes requested):

templates/app/test/routes/root.test.js
Eomm: It is not required by fastify, but by the fastify-cli logic.

for example we hit this line:

https://github.com/fastify/fastify-cli/blob/5654ea1710c5b8e67dd5e73bcfc5349a983bdb10/start.js#L89

that read the package.json and manage the ESM files.

It should not be necessary if we manage simple CJS script

README.md
```diff
@@ -323,6 +323,41 @@ The documentation is rendered using an interactive terminal session that you can
...
+To do so, you can use the this module to load your application and give you the control to write your assertions.
+
+There are two utilities provided:
+
+- `build`: builds your application and returns the `fastify` instance without calling the `listen` method.
+- `listen`: starts your application and returns the `fastify` instance listening on the configured port.
+
+Both of these utilities have the `function(arg, pluginOptions)` parameters:
+
+- `arg`: is a string or a string array within the same arguments passed to the `fastify-cli` command.
```
RafaelGSS, line 338: I would rename it to something more understandable, what o you think about `cliArgs` ?

RafaelGSS (approved):
LGTM.

templates/app/test/helper.js
```diff
@@ -14,13 +14,14 @@ function config () {
 }
 
 // automatically build and tear down our instance
-function build (t) {
-  const app = Fastify()
+async function buildApplication (t) {
```
mcollina, line 17: let's keep calling this `build`.

mcollina (changes requested):
nits, good work

mcollina (approved):
lgtm

PR #485, opened 2022-05-15
Enhance generated VS Code configuration

segevfiner:
Enhanced the generated VS Code configuration for a cleaner experience. Also simplified the `dev` command for the TypeScript template and removed `nodemon`. (P.S. Do we need `ts-node`?)

There is also the `node-terminal` launch configuration type for `launch.json`, but I'm not sure what is the difference in it in contrast to `node`. `pwa-node` is just a leftover from the debugger rewrite from a while back, and `node` is now the same thing.

#### Checklist


mcollina (approved):
lgtm, code looks good!

@fastify/typescript could you take a look?

segevfiner:
Heard of it, but haven't tried it. There are many ways of transpiling TS, so it can get quite opinionated. So what option we use by default should be carefully evaluated, and we might want multiple options.

fox1t:
@segevfiner, I agree 100% with you. We can go with tsc knowing it is slow and generally is not used in the wild, and we might open a new PR in the future to let the user pick the compiler.

mcollina (approved):
lgtm

PR #491, opened 2022-05-27
feat(WIP): rebuild

### LINKED ISSUE #194: Use oclif as cli framework
Hi,
I have very good experiences with https://oclif.io/ it allows to build modular cli applications. All the common tasks like help generation, command/sub-command abstraction, flag, argument parsing or even a plugin architecture is already abstracted in a nice interface.

Some real-world examples:

- https://github.com/heroku/cli
- https://github.com/netlify/cli

It will reduce the burden to get familiar with our cli and make it much more maintainable when the cli gets bigger.

climba03003:
Fixes #194
This rebuild the CLI functionality from scratch using `@oclif/core`.
Currently, I need some helps on the `Github Actions`. Which stale forever when I run the test.

1. ~~For `generate` command, `plugin` still pending to do.~~
2. For `start` command, I removed some of the option since it is duplicated from `export option`.

### Notable Change
1. It is using TypeScript currently, as I am more familiar with the assist of typings.
2. `help` and `readme` is auto generated.
3. `options` now autoload.

### How to Test
1. Clone the Repo
2. Install dependency
3. Build the source using `npm run build`.
4. `node bin/run <command>`

#### Checklist


package.json
```diff
@@ -1,18 +1,28 @@
...
-    "test": "npm run lint && npm run unit && npm run test:typescript",
-    "test:typescript": "tsd templates/plugin && tsc --project templates/app-ts/tsconfig.json && del-cli templates/app-ts/dist"
+    "build": "shx rm -rf dist && tsc -p tsconfig.json",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "npm run lint --fix",
+    "prepack": "npm run build",
+    "pack": "oclif manifest && oclif readme",
+    "postpack": "shx rm -f oclif.manifest.json",
+    "preunit": "npm run pack",
+    "unit": "tap \"test/**/*.test.ts\"",
```
kibertoad, line 23: Any particular reason to go for tap here? E. g. mocha has way wider mainstream usage.

package.json
```diff
@@ -36,58 +46,74 @@
...
+    "eslint-config-standard-with-typescript": "^21.0.1",
+    "eslint-plugin-import": "^2.26.0",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^5.0.0",
+    "oclif": "^3.0.1",
+    "pino-pretty": "^7.6.1",
+    "shx": "^0.3.4",
+    "tap": "^16.2.0",
+    "ts-node": "^10.7.0",
+    "typescript": "~4.4.0"
```
kibertoad, line 85: this is pretty old

src/commands/generate/project.ts
```diff
@@ -0,0 +1,246 @@
...
+      location: flags.location,
+      overwrite: flags.overwrite
+    }))
+
+    if (this.shouldOverwrite) this.questionOverwriteValidate(answer.overwrite)
+
+    Object.assign(answer, await prompt([
+      { type: 'list', name: 'language', message: 'Which language will you use?', default: 'JavaScript', choices: ['JavaScript', 'TypeScript'] },
+      { type: 'list', name: 'lint', message: 'Which linter would you like to use?', default: this.questionLintDefault, choices: this.questionLintChoices },
+      { type: 'list', name: 'test', message: 'Which test framework would you like to use?', default: 'tap', choices: ['tap'] }
```
kibertoad, line 51: I would default to Mocha here

src/commands/generate/project.ts
```diff
@@ -0,0 +1,246 @@
...
+      { type: 'confirm', name: 'overwrite', message: 'The folder already exist. Do you want to overwrite?', default: flags.overwrite ?? false, when: this.questionOverwriteWhen, askAnswered: true }
+    ], {
+      name: args.name,
+      location: flags.location,
+      overwrite: flags.overwrite
+    }))
+
+    if (this.shouldOverwrite) this.questionOverwriteValidate(answer.overwrite)
+
+    Object.assign(answer, await prompt([
```
kibertoad, line 48: having prettier somewhere might be helpful too

src/utils/package-json.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Other Commands
+  if (answer.language === 'JavaScript') {
+    scripts.test = 'tap "test/**/*.test.js"'
+    scripts.start = 'fastify start -l info app.js'
+    scripts.dev = 'fastify start -w -l info -P app.js'
+  }
+  if (answer.language === 'TypeScript') {
+    scripts.test = 'tap "test/**/*.test.ts" --ts'
+    scripts.start = 'fastify start -r ts-node/register -l info app.js'
```
kibertoad, line 62: is this supposed to be prod-ready start? should it be using ts-node then? no command to build & run?

src/utils/package-json.ts
```diff
@@ -0,0 +1,192 @@
...
+    scripts.dev = 'fastify start -r ts-node/register -w -l info -P app.js'
+  }
+  return scripts
+}
+
+export function computeDependencies (_answer: any, pkg: any): { [key: string]: string } {
+  const dependencies: any = {}
+  dependencies['@fastify/autoload'] = pkg.dependencies['@fastify/autoload']
```
kibertoad, line 70: does it make sense to include autoload by default? It makes following structure of the application harder. And I wonder what are the performance implications for large projects.

templates/project/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "module": "CommonJS",
+    "target": "ES2018",
```
kibertoad, line 5: Replacing ES-version based target with Node-version based template import from TypeScript lib might be more convenient for backend

src/utils/package-json.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (answer.language === 'TypeScript') {
+    scripts.test = 'tap "test/**/*.test.ts" --ts'
+    scripts.start = 'fastify start -r ts-node/register -l info app.js'
+    scripts.dev = 'fastify start -r ts-node/register -w -l info -P app.js'
+  }
+  return scripts
+}
+
+export function computeDependencies (_answer: any, pkg: any): { [key: string]: string } {
+  const dependencies: any = {}
```
kibertoad, line 69: I would add some optional, commonly used libs as well, such as fastify-dotenv, fastify-helmet, fastify-healthcheck, fastify-awilix (for DI).

src/utils/start.ts
```diff
@@ -0,0 +1,198 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const inspector = require('inspector')
+    inspector.open(options.debugPort, options.debugAddress)
+  }
+
+  const fastify = Fastify(fastifyOptions)
+
+  await fastify.register(entryPlugin, { prefix: options.prefix })
+
+  await fastify.listen({ port: options.port, host: options.address })
```
kibertoad, line 97: coupling server startup to app definition is not ideal. what if I need app instance for my tests (using app.inject), but I don't want the overhead of starting/stopping server?

src/utils/watch/fastify.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (this as boolean) { console.log(C.red('[fastify-cli] process forced end')) }
+  process.exit(1)
+}
+
+// shared instance
+let fastify: null | FastifyInstance = null
+
+process.on('message', function (event) {
+  console.log('message', event)
+  if (event === GRACEFUL_SHUTDOWN) {
```
kibertoad, line 17: would https://www.npmjs.com/package/close-with-grace be useful here?

src/utils/watch/fastify.ts
```diff
@@ -0,0 +1,55 @@
+import C from 'ansi-colors'
```
kibertoad, line 1: colorette was 3 times faster, last time I checked, and we use it in pino-pretty already.

package.json
climba03003: `tap` is used through out the whole organization. So, it would be the first priority one.
`mocha` and `jest` can be added, but it needs somebody effort.

package.json
climba03003: Fixed in `~4.4.0` just because of `eslint` and `ts-standard`.

src/commands/generate/project.ts
climba03003: `standard` or `eslint` already does it's job.

package.json
kibertoad: There is a difference between internal and external use, no? CLI is not intended to be used primarily internally.

src/utils/package-json.ts
climba03003: Yes, it needs `build` command.
It is doubt-able the needs for running in `node` directly because some of the detail for starting server is hide inside the CLI.

src/utils/package-json.ts
climba03003: > does it make sense to include autoload by default?

For this question, I don't know. I added it just because the previous one use `@fastify/autoload`.
I am the one against using `@fastify/auteload` for beginner.

package.json
kibertoad: ts-standard seems to be barely maintained, are we sure we don't want to rely on something more actively supported? being locked into old TS is really not great

src/utils/start.ts
climba03003: You can look at the current `test` which provide the `function` to create `fastify` instance.

src/utils/package-json.ts
kibertoad: @mcollina Thoughts? Should we drop autoload from default CLI?

src/utils/watch/fastify.ts
climba03003: I would avoid any tools using `"type": "module"`. Which ease the problem of `dependabot` always bumping those package but will never be success.

package.json
climba03003: Using united test framework inside the organization (`fastify`) have some advantages. Developer inside the organization do not need to learn multiple test framework.

> There is a difference between internal and external use, no?

I think it would be a long discussion on test framework topic.

package.json
climba03003: > ts-standard seems to be barely maintained, are we sure we don't want to rely on something more actively supported? 

In here, I would like to share the same style with other packages. I am aware of `ts-standard` is leak of maintaince, but it does the job right now.
Currently, I am just waiting for https://github.com/standard/eslint-config-standard-with-typescript update.

In my projects, I bumped the `typescript` version without problem. I only gives me warning of unsupported `typescript`.
As a tools that will be widely used, I don't think this warning is acceptable and I will wait for the update in `standard` first.

src/utils/watch/fastify.ts
mcollina: @climba03003 the main entry point of colorrette is a cjs file: https://github.com/jorgebucaran/colorette/blob/main/package.json#L5.

src/utils/watch/fastify.ts
mcollina: Definitively a better alternative

templates/project/tsconfig.json
climba03003: The version before, it using `fastify-tsconfig` and it hide all the options from user.
Using the official template face the same problem.

simoneb:
is this PR going anywhere?

climba03003:
> is this PR going anywhere?

Yes, I need some GitHub experts to help me on the GitHub Actions.
It test stale on the GitHub Actions but works on local machine.

Moreover, I need more feedback on how the template should look like.

simoneb:
as far as I can see there are no checks being run in gh actions in the PR. see the checks tab, it says 0. as for the structure, my wild guess is that landing a PR with 150 changes files will be challenging

climba03003:
> as far as I can see there are no checks being run in gh actions in the PR. see the checks tab, it says 0. 

I run it before and I see it stale forever. Then, I removed the ci files.

> as for the structure, my wild guess is that landing a PR with 150 changes files will be challenging

It is a complete rewrite. So, it will be large anyway.

simoneb:
ok. well without seeing the output of ci hard to say what might be the cause

climba03003:
> ok. well without seeing the output of ci hard to say what might be the cause

Here is the clean repo.
https://github.com/kaka-repo/fastify-cli

You can check the CI in here. I have push the CI file again for testing.
https://github.com/kaka-repo/fastify-cli/runs/8204969713

simoneb:
that's not a hang. the operation was canceled because another job failed https://github.com/kaka-repo/fastify-cli/runs/8204970017?check_suite_focus=true

climba03003:
> that's not a hang. the operation was canceled because another job failed https://github.com/kaka-repo/fastify-cli/runs/8204970017?check_suite_focus=true

Here is the hanging one https://github.com/kaka-repo/fastify-cli/runs/8205522639?check_suite_focus=true

sinedied:
I know it's a bit late for this given the work already done, but I'm a bit sad to see this change as the CLI was previously quite fast to install and lightweight: not trying to start a debate regarding features here, but `oclif` is quite heavy and brings in 92 (!) dependencies just on its core: https://npmgraph.js.org/?q=%40oclif%2Fcore
For the sake of comparison, `commander` brings similar benefits than `oclif`, with no added dependencies: https://npmgraph.js.org/?q=commander

Just sharing my maintainer experience here, but besides the bloat, added dependencies (even transitive ones) adds burden to the maintainers because of frequent updates and potential security issues.

arthurfiorette:
Hey @climba03003, @sinedied. Any updates here?

Oclif would be an amazing feature! We know that oclif can be "bloated", however, the DX features it brings make it so valuable. Autocompletion of commands, ability to add plugins, cli validation and many other things that previously would have to be done manually will now be generated by oclif for us.

This will drastically reduce the amount of bugs to fix and code we would have to write.

jean-michelet:
I think we should give it another try, maybe starting a new clean PR.

These utilities have proved their worth and will be well received by people migrating from other languages who use them a lot.

The main problem is CI, right?

climba03003:
> The main problem is CI, right?

Yes, at least from last try. It never been able to works on Github Actions.
But works properly on local machines.

If anyone want a try or another PR.
He / She should use https://nodejs.org/docs/latest/api/util.html#utilparseargsconfig instead.
Then, in combine with `inquirer`

jean-michelet:
Should we start by increasing code coverage or focus on this feature first?
Seems like it would break some stuff, we could focus on coverage later?

climba03003:
Focus on implementation first, before caring on the test coverage.

climba03003:
Replaced by #743

Issue #363, opened 2021-05-08
Add minimal eslint and Prettier configurations


veritem:
It would be great if for the generated code you added minimal eslint and prettier configurations as optional for people that need them in their work flow

mcollina:
We did not do this on purpose as we could not really agree on these topics between ourselves.

What are the other folks in @fastify/fastify think? I'm happy to go with the majority.

(As an example, I dislike any use of prettier).

zekth:
I think that have to be an option in the creation of a project like `Would you want to have code style configurations?`.
EG: in most of my projects i use eslint + editorconfig + prettier because of dayjob habits; but having those forced in a cli like in `vue-cli` or any other is a bit too intrusive to me. I personnaly don't mind of the code style as the scripts in `package.json` are implemented like `lint` and `lint:fix`

Eomm:
I agree with the optional flag to turn on this feature

Someone else who prefer another setup (es: `standardjs`) shall submit a PR 👍🏽

meehawk:
@mcollina I would love to work on this if it is still open. Trying to get more contributions to fastify in 😁

UltiRequiem:
Heey, this will be super useful 😩

Mazuh:
just opened a PR about it... I didn't want to open space for multiple eslint choices, I just automated (if user desires it) what was already suggested by the readme file, cause I'm not wanting to start any kind of discussion about code style choices to support nor anything like that right now.

PR #505, opened 2022-06-13
feat: lint option for javascript template

Mazuh:
resolves https://github.com/fastify/fastify-cli/issues/363

#### Checklist

- [x] run `npm run unit`

for typescript, see: https://github.com/fastify/fastify-cli/pull/512

### DIFF
--- a/args.js
+++ b/args.js
@@ -12,7 +12,7 @@ module.exports = function parseArgs (args) {
       'populate--': true
     },
     number: ['port', 'inspect-port', 'body-limit', 'plugin-timeout'],
-    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require'],
+    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require', 'standardlint'],
     boolean: ['pretty-logs', 'options', 'watch', 'verbose-watch', 'debug'],
     envPrefix: 'FASTIFY_',
     alias: {
@@ -40,7 +40,8 @@ module.exports = function parseArgs (args) {
       debugPort: 9320,
       options: false,
       'plugin-timeout': 10 * 1000, // everything should load in 10 seconds
-      lang: 'js'
+      lang: 'js',
+      standardlint: false
     }
   })
 
--- a/args.js
+++ b/args.js
@@ -12,8 +12,8 @@ module.exports = function parseArgs (args) {
       'populate--': true
     },
     number: ['port', 'inspect-port', 'body-limit', 'plugin-timeout'],
-    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require', 'standardlint'],
-    boolean: ['pretty-logs', 'options', 'watch', 'verbose-watch', 'debug'],
+    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require'],
+    boolean: ['pretty-logs', 'options', 'watch', 'verbose-watch', 'debug', 'standardlint'],
     envPrefix: 'FASTIFY_',
     alias: {
       port: ['p'],
--- a/generate.js
+++ b/generate.js
@@ -38,6 +38,10 @@ const javascriptTemplate = {
     log('debug', `run '${chalk.bold('npm start')}' to start the application`)
     log('debug', `run '${chalk.bold('npm run dev')}' to start the application with pino-colada pretty logging (not suitable for production)`)
     log('debug', `run '${chalk.bold('npm test')}' to execute the unit tests`)
+
+    if (Object.keys(pkg.scripts).includes('lint')) {
+      log('debug', `run '${chalk.bold('npm lint')}' to run linter and fix code style issues`)
+    }
   }
 }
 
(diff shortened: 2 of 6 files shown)


test/generate.test.js
```diff
@@ -159,6 +171,23 @@ function define (t) {
     })
   }
 
+  function verifyStandardlintInPkg (t) {
+    return new Promise((resolve, reject) => {
+      const pkgFile = path.join(workdir, 'package.json')
+
+      readFile(pkgFile, function (err, data) {
```
mcollina, line 178: Could you please use `fs.promises.readFile()`.

mcollina (changes requested):

test/generate.test.js
Mazuh: sure, I just tried to follow the existing code pattern, but it's fixed with what you asked now.

mcollina (approved):
lgtm

climba03003 (changes requested):
I would change the option to `lint` which can be extendable in future.

climba03003:
> IMHO people that need something beyond the standard linter will probably do that with their own custom *rc files (.prettierc, .eslintrc), each one requiring extra dependencies (every eslint project has tons of different plugins listed on devdependencies), at least that's my experience, not to mention that eslint/prettier CLIs are already very customizable by themselves.

It is a `string` option which mean it can be extended like `--lint=standard`, `--lint=eslint`, etc.

args.js
```diff
@@ -12,7 +12,7 @@ module.exports = function parseArgs (args) {
       'populate--': true
     },
     number: ['port', 'inspect-port', 'body-limit', 'plugin-timeout'],
-    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require'],
+    string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require', 'standardlint'],
```
climba03003, line 15: At least I see you declare it as string option.

Mazuh: Good point. That's because I started to code thinking of it to be a string indeed, then I thought "if someone really tries to put more options here, it will be annoying AF to decide and maintain each non-standard option" and so I changed my mind.

but, again, I'm just waiting for some agreed final design choice from all reviewers, it should be an easy code change anyway.

generate.js
```diff
@@ -149,7 +153,25 @@ function cli (args) {
...
+  if (opts.lang === 'ts' || opts.lang === 'typescript') {
+    template = typescriptTemplate
+  } else {
+    template = { ...javascriptTemplate }
+
+    if (opts.standardlint) {
+      template.scripts = {
+        ...template.scripts,
+        pretest: 'standard',
+        lint: 'standard --fix'
```
Eomm, line 166: In this org we use:

```
lint -> run the linter
lint:fix -> run the linter's fixes
```

Eomm (commented):
> tldr: I tried to avoid a lot of headache for maintainers.

**+1**

I must say that it would be awesome to think of a "pluggable" cli (or a yeomen generator? 🤔 )

mcollina (approved):
lgtm

generate.js
Mazuh: @Eomm, I prefer such pattern too, but this is for the template, not the org project itself, and for the typescript template it's already just `lint` instead of `lint:fix`, and just `lint` is already what's recommended by the readme right now.

so here I ask two questions:
- should I proceed changing it to `lint:fix`? 
- if yes, then should I update typescript template too (along with its docs) (and readme suggestion)?

climba03003:
> I must say that it would be awesome to think of a "pluggable" cli (or a yeomen generator? 🤔 )

@Eomm I recommend to check this PR https://github.com/fastify/fastify-cli/pull/491

Mazuh:
@climba03003 @Eomm any feedback on this? should I assume that this PR should be close (it looks like the issue is being worked on in another PR or something)?

generate.js
Eomm: I reread the readme and I would not change it as it should be already discussed in past

Eomm (approved):

Eomm (approved):
LGTM when ci green

simoneb:
@Mazuh let's figure out the windows problem first

Mazuh:
it was indeed just a testing issue, the `rimraf.sync(...)` that runs before each test behaves a little bit differently on windows, all others test cases were running fine but my little one stepped on a race condition and I was trying to generate the project in a busy directory.

now I have my ubuntu and my windows with `npm t` returning 0 (ok), so the same should happen with the pipeline, I'm going back to my macbook to confirm that everything is ok there too.

Captura de tela 2022-06-21 081851

mcollina:
@simoneb you good to land this?

help/generate.txt
```diff
@@ -11,3 +11,6 @@ OPTS
 
   --lang=ts, --lang=typescript
       use the TypeScript template
+
+  --standardlint
```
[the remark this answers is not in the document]
simoneb: also, any reason why we don't do it for the ts template?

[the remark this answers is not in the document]
Mazuh: yes, tests will pass but it won't work so easily like js support, I opened another PR for further discussions: https://github.com/fastify/fastify-cli/pull/512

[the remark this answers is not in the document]
Mazuh: and about the option name, it's a consequence of a few discussions above and the conclusion is to **narrow** this feature to a simple yes-or-no support, while some other future feature (afaik, that already has a PR in WIP) will add a more broad support to other linter options.

simoneb (approved):

simoneb:
@mcollina yes LGTM

Issue #457, opened 2022-02-04
Typescript tests in subfolder of routes are not executed


braekevelt:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.27.1

### Plugin version

2.15.0

### Node.js version

14.18.2

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20.04

### Description

If you move `app/test/routes/example.test.ts` to `app/test/routes/subfolder/example.test.ts`, it is not longer executed when running `npm run test`. This can be fixed by changing `tap --ts test/**/*.test.ts` to `tap --ts test/plugins test/routes`.

### Steps to Reproduce

Generate an app using `npx fastify-cli generate app --lang=ts`.
Run `npm run test`.
Observe that 3 tests are executed.
![image](https://user-images.githubusercontent.com/28069994/152558577-5e2ec733-9caa-4a11-870e-8b01a6a7a6c2.png)

Create a folder `subfolder` inside `app/test/routes`.
Move `app/test/routes/example.test.ts` to `app/test/routes/subfolder/example.test.ts`.
Run `npm run test`.
Observe that only 2 tests are executed.
![image](https://user-images.githubusercontent.com/28069994/152558688-ad401220-d5b2-439b-a215-4e07a080a836.png)

### Expected Behavior

Moving tests to subfolders should not prevent their execution.

mcollina:
Would you like to send a Pull Request to address this issue?

KXLAA:
I could try to sort this out if @braekevelt doesn't want to

braekevelt:
@KXLAA you can make the Pull Request if you want to. Thank you!

is2ei:
I guess we don't need to specify each subdirectory.

```
tap --ts test/plugins test/routes
```

The command below is enough.

```
tap --ts
```

https://node-tap.org/docs/cli/
> By default, tap will search for all files ending in
> .ts, .tsx, .js, .jsx, .cjs, or .mjs, in a top-level
> folder named test, tests, or __tests__, or any file
> ending in '.spec.' or '.test.' before a supported
> extension, or a top-level file named
> 'test.(js,jsx,...)' or 'tests.(js,jsx,...)'

is2ei:
I found `fastify-cli` uses `minimatch` for unit testing but `tap` uses `glob`.

[...]

PR #506, opened 2022-06-14
Fix-457: TS Testing pattern fix

nnorbert:
#### Checklist

### DIFF
--- a/generate-plugin.js
+++ b/generate-plugin.js
@@ -23,7 +23,7 @@ const pluginTemplate = {
     'lint:typescript': 'ts-standard',
     test: 'npm run lint && npm run unit && npm run test:typescript',
     'test:typescript': 'tsd',
-    unit: 'tap test/**/*.test.js'
+    unit: 'tap "test/**/*.test.js"'
   },
   dependencies: {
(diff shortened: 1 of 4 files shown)


simoneb:
can you please describe what the PR does and why it works?

climba03003:
It works because the current command is using `shell` glob.
With the quote, it use `glob` package.

This problem is actually depends on system `shell`.
We face the similar problem before in `fastify-swagger` but I forgotten which PR is it.

nnorbert:
The problem was the following, NPM uses **sh** as the shell to run commands which one does not understand the ** glob pattern, it will match only one level. To protect your pattern from being interpreted by sh, simply pass the pattern as a string parameter. If you do so, it will be evaluated by a JavaScript implementation of globs.
Reference: [https://stackoverflow.com/questions/32017169/npm-glob-pattern-not-matching-subdirectories](https://stackoverflow.com/questions/32017169/npm-glob-pattern-not-matching-subdirectories)

simoneb (approved):

simoneb:
Closes #457

Issue #513, opened 2022-06-25
Generate typescript app not starting


kamo93:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.0.0

### Plugin version

3.0.0

### Node.js version

14.17.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 20.04.4

### Description

When trying to start a just generate app using typescript the browser shows 
![image](https://user-images.githubusercontent.com/18643841/175788034-782f4dcb-f97e-4f7f-8e78-e01d4286e396.png)

When trying to generate and start a js fastify app everything work fine  but with typescript doesn't start.

This is what  it shows the console.
![image](https://user-images.githubusercontent.com/18643841/175788160-b27ab9f2-b672-437f-a58a-32de148c701f.png)

### Steps to Reproduce

Generate an app, and trying to start it.
```
fastify generate <app-name> --lang=ts

npm install

npm start dev
```

### Expected Behavior

Start the app normally

kamo93:
The only work around that I found was to run `npm run watch:ts` and `npm run dev:start` on different terminals. But sometimes while the new dist/app.js gets build the server response unavailable service just for a moment.

mcollina:
Where did you see you should run `npm start dev`?  Running `npm start` works just fine.

I can see the problem you are describing with `npm run dev`. Is this correct?

mcollina:
@fastify/typescript could you take a look here? I don't understand why `npm run dev` is not really working.

kamo93:
Sorry my mistake. The problem occurs with `npm run dev`

BigGillyStyle:
FWIW...I also am seeing this same problem in this local config:

MacOS: Monterey 12.4
Node: 16.15.1
NPM: 8.13.1

```shell
$ npm ls
├── @fastify/autoload@5.0.0
├── @fastify/sensible@5.1.0
├── @types/node@16.11.41
├── @types/tap@15.0.7
├── concurrently@7.2.2
├── fastify-cli@4.2.0
├── fastify-plugin@3.0.1
├── fastify-tsconfig@1.0.1
├── fastify@4.1.0
├── tap@16.3.0
├── ts-node@10.8.1
└── typescript@4.7.4
```

mcollina:
I have no clue why this is happening and not really much time to look into it right now. A PR that fixes this would be awesome.

[...]

PR #511, opened 2022-06-20
fix unreachable if-block

Uzlopak:
Resolves #510

PR #515, opened 2022-06-28
Handle 'exit' of child_process

BigGillyStyle:
Fixes https://github.com/fastify/fastify-cli/issues/513.  Reverts change made in https://github.com/fastify/fastify-cli/pull/511 and subsequently release 4.2.0 of fastify-cli that prevented TS-generated projects to work properly when launched from `npm run dev` script.

#### Checklist

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -48,7 +48,7 @@ const watch = function (args, ignoreWatch, verboseWatch) {
     })
 
     _child.on('exit', function (code, signal) {
-      if (!code) {
+      if (!code === 0) {
         stop()
       }
       if (childs.length === 0 && !allStop) {
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -48,9 +48,6 @@ const watch = function (args, ignoreWatch, verboseWatch) {
     })
 
     _child.on('exit', function (code, signal) {
-      if (!code === 0) {
-        stop()
-      }
       if (childs.length === 0 && !allStop) {
         childs.push(run('restart'))
       }
--- a/test/start.test.js
+++ b/test/start.test.js
@@ -533,8 +533,7 @@ test('should start the server listening on 0.0.0.0 when running in docker', asyn
   t.pass('server closed')
 })
 
-// TODO remove skip
-test('should start the server with watch options that the child process restart when directory changed', { skip: true }, async (t) => {
+test('should start the server with watch options that the child process restart when directory changed', async (t) => {
   t.plan(4)
   const tmpjs = path.resolve(baseFilename + '.js')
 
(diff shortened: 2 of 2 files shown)


BigGillyStyle:
This is still a work in progress.  I am working on extra verification of this fix locally as well as seeing about writing unit tests.

lib/watch/index.js
```diff
@@ -48,7 +48,7 @@ const watch = function (args, ignoreWatch, verboseWatch) {
     })
 
     _child.on('exit', function (code, signal) {
-      if (!code) {
+      if (!code === 0) {
```
Eomm, line 51: Does the `!` coerce the `code` to a boolean?

mcollina, line 51:
```suggestion
      if (code === 1) {
```

Would this work the same?

mcollina (approved):
lgtm

BigGillyStyle:
@Eomm and @mcollina Refer to my comments here: https://github.com/fastify/fastify-cli/issues/513#issuecomment-1169847178

Bottom line...I created a new project using fastify-cli 4.2.0 using both the JS and TS versions.  I started the app to verify I could successfully access http://127.0.0.1:3000/, and then I made a simple code change to the `root` route in the app.  This PR originally reverted the code in question to "as was" (4.1.1 and prior)...but then I realized that JavaScript negation would never result in a `0` value...so it was simply unreachable code.  So my latest change just removes that code block altogether, and I put that modified code through the same manual "test suite".

mcollina:
Do you think it's feasible to write a test to check if this could regress at all?

lib/watch/index.js
bcomnes: Seems like this would be more in line with what is intended with the original code, and it also makes sense (not knowing the broader context here).

bcomnes:
> Do you think it's feasible to write a test to check if this could regress at all?


I can take a look at doing that.

BigGillyStyle:
> Do you think it's feasible to write a test to check if this could regress at all?

Sorry...work was all-consuming towards the end of last week.  I'll also see what I can do about writing tests.

BigGillyStyle:
I re-enabled some previously-skipped tests, as they seem to target this scenario.  I ran them with the "broken code" (before my patch), and _occasionally_ the tests would just hang.  I tried cleaning out the `test/fixtures/*.js` files, `test/workdir` folder, and `test/workdir-with-lint` folder between tests, but that did not seem to correlate with why the tests would occasionally hang...and sometimes wouldn't 🤷   I never did experience a hang with the patch from this PR, but I can't say that I'm 100% confident that (1) the tests won't hang again and/or (2) this properly tests all the right scenarios for this issue.

BigGillyStyle:
@bcomnes has all of this covered in https://github.com/fastify/fastify-cli/pull/519 and has more work and test fixes there, so this PR is being closed.  I'm glad we could all work together on a solution here.

PR #519, opened 2022-07-03
Fix watch regression

### LINKED ISSUE #513: Generate typescript app not starting

### LINKED ISSUE #518: Watch never restarts the server after multiple watch events (v4.2.0)
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.2.0

### CLI version

4.2.0

### Node.js version

18.4.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.4

### Description

After updating to fastify-cli 4.2.0 from 4.1.1, it seems as though modifying watched files can cause the fastify-cli server to stop the server and never restart. Killing the process after this happens, the following error is observed sometimes:

```console
% Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
[fastify-cli] app crashed - waiting for file changes before starting...
[1656790027711] INFO (42069 on MacBook-Pro.bec.local): Server listening at http://0.0.0.0:3000
```

A samble of verbose logging prior to the server not responding/restarting looks like this:

```
[fastify-cli] watch - 'addDir' occurred on 'public/hooks'
[fastify-cli] watch - 'addDir' occurred on 'public/lib'
[fastify-cli] watch - 'addDir' occurred on 'public/login'
[fastify-cli] watch - 'addDir' occurred on 'public/logout'
[fastify-cli] watch - 'addDir' occurred on 'public/register'
[fastify-cli] watch - 'addDir' occurred on 'public/resources'
[fastify-cli] watch - 'addDir' occurred on 'public/tags'
```

After these messages print, the server is no longe accepting requests. 

### Steps to Reproduce

Start a fastify-cli server in watch mode running fastify-cli 4.2.0 and then trigger some file changes. A `mkdir -p` works well. The server restarts if there are only one or two file watch actions at a time. It will not restart if there are 3-4 actions at a time. Downgrading the CLI to 4.1.1 restore a functional watch mode. 

[Sample repo](https://github.com/ballpit/fastify-cli-bug)

### Expected Behavior

Server should eventually accept requests after files change in watch mode.

bcomnes:
This fixes tests that were not passing on master, improved tests to better match behavior and also fix the regression introduced in https://github.com/fastify/fastify-cli/pull/511 described in https://github.com/fastify/fastify-cli/issues/518 and https://github.com/fastify/fastify-cli/issues/513 and https://github.com/fastify/fastify-cli/pull/515.

I will push up a regression test asap, but I want to validate these changes pass in CI here at this point. 

Fixes #513.
Fixes #518.

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -48,9 +48,6 @@ const watch = function (args, ignoreWatch, verboseWatch) {
     })
 
     _child.on('exit', function (code, signal) {
-      if (!code) {
-        stop()
-      }
       if (childs.length === 0 && !allStop) {
         childs.push(run('restart'))
       }
--- a/test/start.test.js
+++ b/test/start.test.js
@@ -604,7 +604,7 @@ test('should start the server with watch and verbose-watch options that the chil
   t.ok(spy.args.length > 0, 'should print a console message on file update')
 })
 
-test('should reload the env on restart when watching', async (t) => {
+test('should reload the env on restart when watching', { skip: process.platform === 'win32' }, async (t) => {
   const testdir = t.testdir({
     '.env': 'GREETING=world',
     'plugin.js': await readFile(path.join(__dirname, '../examples/plugin-with-env.js'))
--- a/test/start.test.js
+++ b/test/start.test.js
@@ -251,8 +251,10 @@ test('should error with a good timeout value', async t => {
     }
   })
 
+  const port = getPort()
+
   try {
-    const argv = ['-p', '3040', '-T', '100', './test/data/timeout-plugin.js']
+    const argv = ['-p', port, '-T', '100', './test/data/timeout-plugin.js']
     await start.start(argv)
   } catch (err) {
     t.equal(err.code, 'AVV_ERR_READY_TIMEOUT')
@@ -533,13 +535,14 @@ test('should start the server listening on 0.0.0.0 when running in docker', asyn
   t.pass('server closed')
 })
 
-// TODO remove skip
-test('should start the server with watch options that the child process restart when directory changed', { skip: true }, async (t) => {
-  t.plan(4)
+test('should start the server with watch options that the child process restart when directory changed', { skip: process.platform === 'win32' }, async (t) => {
+  t.plan(3)
   const tmpjs = path.resolve(baseFilename + '.js')
 
+  const port = getPort()
+
   await writeFile(tmpjs, 'hello world')
-  const argv = ['-p', '4042', '-w', './examples/plugin.js']
+  const argv = ['-p', port, '-w', './examples/plugin.js']
   const fastifyEmitter = await start.start(argv)
 
   t.teardown(async () => {
@@ -548,9 +551,6 @@ test('should start the server with watch options that the child process restart 
     }
     await fastifyEmitter.stop()
   })
-
-  await once(fastifyEmitter, 'start')
-  t.pass('should receive start event')
 
   await once(fastifyEmitter, 'ready')
   t.pass('should receive ready event')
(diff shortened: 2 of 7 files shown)


mcollina (approved):
lgtm

lib/watch/index.js
```diff
@@ -48,7 +48,7 @@ const watch = function (args, ignoreWatch, verboseWatch) {
     })
 
     _child.on('exit', function (code, signal) {
-      if (!code) {
+      if (code > 0) {
```
Eomm, line 51: Could you describe the intent here?

`!0` would be `true`, while `0 > 0` is `false` so I don't get why this inversion

mcollina:
CI is failing

lib/watch/index.js
mcollina: We can delete this code block

bcomnes: It seems like the idea was that if code can either be null or an exit code integer. In the case of a > 0 code, the watcher should stop. In the case of null or 0, restart. It was ignoring > 0 codes before, mostly resulting in desired behavior. 

I can also just remove the block.

bcomnes:
Darn looking into it.

lib/watch/index.js
BigGillyStyle: While working through Fastify CLI generated JS and TS projects, I _did_ find after changing an app file that the `code` would be `1` in each case I tried...leading to the `stop()` call in line 52 here.  However, removing the `if` block altogether seemed to produce identical behavior.  I don't understand enough of what is going on to determine which option is correct.  I don't know if either choice might lead to dangling processes or not.

BigGillyStyle: I guess I should also note that outside the scenario of the app restarting due to a source file change, I otherwise observed that the `code` would be `0` or `null`.

bcomnes: Ah ok. Sounds like we should stick with the original behavior and just remove the switch block all together then.  Sorry to add more noise, I didn't pick up on that caveat. 

If I find some time today I'll see if I can investigate the tests failing. 

I noticed that the (currently disabled) watch tests were disabled in ci due to lack of chokidar support for Linux (or windows)? I wonder, if we can restore these tests, could we run them in the macOS image?

bcomnes: Cherry picked your patch that does this.

bcomnes:
Oooookie dokie. This should be good to go, CI passing in my fork:

https://github.com/bcomnes/fastify-cli/actions/runs/2606233931

I also confirmed turning on these tests would have caught the regression in #511.

I also cherry picked @BigGillyStyle's patch so he's on the commit list.

test/start.test.js
```diff
@@ -591,9 +592,6 @@ test('should start the server with watch and verbose-watch options that the chil
     await fastifyEmitter.stop()
   })
 
-  await once(fastifyEmitter, 'start')
-  t.pass('should receive start event')
-
```
bcomnes, line 596: I'm not sure why, but these start events never fire anymore.

BigGillyStyle:
> I also cherry picked @BigGillyStyle's patch so he's on the commit list.

Heh heh...thanks for doing that, but no worries either.  Mainly hoping I can help with the solution and not add noise 😄  I'll go ahead and close my PR.  Thanks @bcomnes and @mcollina

bcomnes:
Darn another windows timeout. Let me increase the timeout on that file.

bcomnes:
Sorry, maybe try one more time. I increase the timeout on tests. Seems as though the windows tests fail randomly.

Issue #401, opened 2021-07-21
Config File Option


galvez:
Would be nice to be able to pass `-c config.js` in the command line, a file that would export an object with this signature:

```
{
  port, 
  address, 
  socket, 
  require, 
  logLevel, 
  debug, 
  debugPort, 
  debugHost, 
  prettyLogs, 
  watch, 
  ignoreWatch, 
  verboseWatch, 
  options, 
  prefix,
  pluginTimeout,
  bodyLimit,
}
```

That is, a 1:1 mapping to all options configurable via environment variables.

Would you consider merging a PR implementing this feature?

mcollina:
Sure thing!

wlinna:
Is there a workaround for this? I am new to fastify, and I'd like to set trustProxy to true, but I have no idea how to do that when using fastify-cli.  #497  says that there isn't any way to set it, but is it true?

Eomm:
Right now we have the `--options` flag that reads the JSON config from the app itself: https://github.com/fastify/fastify-cli#start

nooreldeensalah:
I'm interested in working on this issue, but I wanted to inquire about a specific detail.

Let's say, for example, there's a `config.js` file with the following options

```javascript
module.exports = {
    port: 5000,
    prettyLogs: true
}
```

And the user while starting the application passed both `--config=config.js` and `--port=6000`, which port number should be given the priority, or should this return an error?

PR #522, opened 2022-07-04
feat: add support for configuration file

nooreldeensalah:
In this PR, I've added `--config` or `-c` command line argument that takes a configuration file containing options related to `fastify-cli`.

If there are options listed both as command line arguments and in the config file, the command line argument options are prioritized.

This PR addresses #401

#### Checklist


nooreldeensalah:
https://github.com/nooreldeensalah/fastify-cli/blob/2c53c383b793c3b47d6c2183d7da2c55ab6d25cc/args.js#L89-L99

Concerning this part, I've thought of using the `detailed` function of `yargs-parser` and then get the default values, then use a regex to convert from kebab-case to camelCase, and modify `args.js` accordingly.

Do you think this will be a good idea?

nooreldeensalah:
>If there are options listed as both command line arguments and in the config file, the command line argument options are prioritized.

A very specific corner-case to this scenario would be, if a default argument being passed as a command line argument with the default value, but the case doesn't make much sense. Here is an example to clarify what I mean.

For example, the default value of `debugPort` is `9320`.

```javascript
\\ config.js
module.exports = {
 debugPort: 5000
}
```
Along with `--debugPort=9320`, in this case `debugPort` will be equal to `5000` (i.e., the config file parameter is prioritized), but that's a very specific corner case which doesn't make sense, like one wouldn't pass a command line argument for it to be the same as the default one.

climba03003 (changes requested):
I think the argument passed through cli should take the highest priority.
Since adding argument in cli is more explicit and file can be the user default.

mcollina (approved):
lgtm

README.md
```diff
@@ -127,10 +127,11 @@ This works with a `.js` extension if you are using Node.js >= 14 and the nearest
...
-You can pass the following options via CLI arguments. Every option has a corresponding environment variable:
+You can pass the following options via CLI arguments. You can also use `--config` or `-c` flag to pass a configuration file that exports all the properties listed below in camelCase convention. In case of collision (i.e., An argument existing in both the configuration file and as a command-line argument, the command-line argument is given the priority). Every option has a corresponding environment variable:
 
 | Description                                                                                                                             | Short command | Full command       | Environment variable     |
 | --------------------------------------------------------------------------------------------------------------------------------------- | ------------- | ------------------ | ------------------------ |
+| Path to configuration file that can be used to manage the options listed below                                                                                                   | `-config`          | `--config`           | `FASTIFY_CONFIG or CONFIG`   |
```
mcollina, line 134:
```suggestion
| Path to configuration file that can be used to manage the options listed below                                                                                                   | `-c`          | `--config`           | `FASTIFY_CONFIG or CONFIG`   |
```

mcollina (changes requested):

nooreldeensalah:
> I think the argument passed through cli should take the highest priority. Since adding argument in cli is more explicit and file can be the user default.

Yes I do agree, and this is the implemented behavior, but a corner case would be as stated in my second comment (the explicit argument has the same value as the default one), there is no way to tell that arguments were passed explicitly in that case.

But in the default cases such as `config.js` containing `port=5000` and `--port=6000`, port will end up being `6000`

nooreldeensalah:
> You can remove the `default` passed to `yargs-parser`.

Thanks for your suggestion, I went with this approach, and now the options are parsed according to this order from lower priority to higher priority (Default Arguments → Config file options → Command line arguments). The logic is much simpler now.

I also wanted to point out to a regression that was caused by removing the default arguments, that caused the test ([should parse env values correctly](https://github.com/fastify/fastify-cli/blob/742615d390057a2aa146e78b4a0fa553966310c6/test/args.test.js#L105)) to fail, where it was `expected: true` and `actual : 'true'`. After adding an alias `-V` for `--verbose-watch`, the failing test passed correctly.

Please review the changes, and if everything is alright, I'll rebase to autosquash the fixup commits  .

climba03003 (approved):
LGTM.

mcollina (approved):
lgtm

PR #523, opened 2022-07-08
feat: support exporting server options using `--options` on ESM

### LINKED ISSUE #411: Support ESM for exporting server `--options`
It would be great if ESM projects could `export const options` when using the `--options` flag.

Like this:

```js
export default async (fastify, opts) => {
  // ...
}

export const options = {
  maxParamLength: 200,
}
```

For now, this seems to work:

```js
async function server(fastify, opts) {
  // ...
}

server.options = {
  maxParamLength: 200,
}

export default server
```

Not the end of the world, but I did spin my wheels for a bit thinking `export const options` would just work, since ESM is supported otherwise.

nooreldeensalah:
This PR resolves #411
#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,21 @@ module.exports.options = {
 }
 ```
 
+And if you are using EcmaScript Module format:
+
+```javascript
+export default async function plugin (fastify, options) {
+  // Both `/foo` and `/foo/` are registered
+  fastify.get('/foo/', async function (req, reply) {
+    return 'foo'
+  })
+}
+
+export const options = {
+  ignoreTrailingSlash: true
+}
+```
+
 If you want to use custom options for your plugin, just add them after the `--` terminator.
 
 ```js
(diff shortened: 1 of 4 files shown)


nooreldeensalah:
I still haven't added unit-tests, I'll see what would be the best way to test for these functionality then work on the tests, until now, I'll mark the PR as a draft

nooreldeensalah:
I've added tests, and updated the documentation

mcollina (approved):
lgtm

nooreldeensalah:
I wanted to point out that I found a broken link in the routes' documentation in the generated `app` template, specifically here:

https://github.com/fastify/fastify-cli/blob/dd476a793a637da367d278a5041395ef1b909949/templates/app/routes/README.md?plain=1#L26-L27

Where the link should be https://www.fastify.io/docs/latest/Reference/Routes/#promise-resolution (`/Reference`) is missing

Would you like to fix this in a separate PR? Or should I make a commit fixing it here?

mcollina:
> Would you like to have this fixed in a separate PR? Or should I make a commit fixing it here?

A fresh PR would be awesome, I'm shipping this now.

Issue #531, opened 2022-08-04
Improvements to graceful termination strategy suggested by templates


pclalv:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Templates using close-with-grace should use a delay should be greater 500ms

as of the time of writing this, the eject template uses close-with-grace with a delay of 500ms:  https://github.com/fastify/fastify-cli/blob/f4453bfff028844b2e0ce9930d4055d08fe018d5/templates/eject/server.js#L22

cross-referencing this with [the close-with-grace source code](https://github.com/mcollina/close-with-grace/blob/c97f7a29da92d7c7bdfd46ca7a2bec5a096ae65a/index.js#L92-L97), this means that close-with-grace will forcibly exit its current NodeJS process after waiting just 500ms for its callback to complete.

I discovered that a fastify app using this delay is effectively given only 500ms to complete any requests that it'd already started responding to at the time that `SIGTERM` or `SIGINT` was received. so, a request that takes one second to complete will run for half a second and then be cut off, and the requester will see a 50x response. this can be validated this with a simple route that sleeps for a few seconds before returning a response.

that seems a little unreasonably short to me, given that platforms like [Kubernetes][k8s-pod-lifecycle] and [Heroku][heroku-app-restart] first send `SIGTERM`, and then, after a grace period of 30 seconds (by default), `SIGKILL`. why should this default to terminating itself after a mere 500ms when Heroku and Kubernetes both afford orders of magnitude more time for a process to gracefully terminate? does it make sense to use this default, given that many users of this template are liable to use the value of 500ms blindly? actually, that's not a fact, but I bet you it's true :) and users who know better will know better - it's the users who don't know better that we might want to be more considerate of.

### Avoid forcibly exiting the parent NodeJS process

close-with-grace ultimately calls `exit` on its NodeJS process, either when its delay has run out, or when its callback has succeeded or failed, whichever comes first. I find this a little presumptuous - if by an innocent mistake something needing graceful termination isn't being handled in the close-with-grace callback, it'll just be killed by the `exit`. I might want to know if such a condition were affecting my app.

[...]

PR #537, opened 2022-08-13
Add .env conf to graceful termination strategy suggested by templates

avanelli:
### Description 
- Add use of .env FASTIFY_CLOSE_GRACE_DELAY for closeWithGrace in templates eject/server.js
- Add --close-grace-delay parameter

Related to Issue  #531

#### Checklist


mcollina (changes requested):
Could you do the same in `start.js`?

Docs and tests are missing too

avanelli:
> Could you do the same in `start.js`?

Adding to `start.js` should be helpful to add it also as CLI parameter `--close-grace-delay`. Is it ok? 
I can add the new parameter tests in `args.test.js`.
If we need some specific test about `closeWithGrace` timings in `start.test.js`, I will need some indication or example.

> Docs and tests are missing too

About templates eject/server.js and eject-ts/server.ts, I didn't find any relevant tests or docs.
I only changed the server.js code in README.md to reflect the template.

mcollina:
> Adding to start.js should be helpful to add it also as CLI parameter --close-grace-delay. Is it ok?


That would be ok!

Cheers

mcollina (approved):
lgtm

Fdawgs (approved):

Issue #558, opened 2022-10-14
.env values override existing env vars when using --watch


tandibar:
If you start a fastify app with the fastify cli and the watch parameter, the environment variable precedence is changed.

## Example

So if you have a `.env` file with
    
    MY_ENV_VAR=foo

and you run
    
    MY_ENV_VAR=bar fastify start app.js

then `process.env.MY_ENV_VAR` will be 'bar'. Perfect! But if you run it with the watch parameter like

    MY_ENV_VAR=bar fastify start --watch app.js

then `process.env.MY_ENV_VAR` will be 'foo'.

## Suggestion

I think it has something to do with [this line of code](https://github.com/fastify/fastify-cli/blob/18432088e60a2cb8eef7f323da2d0e740536128a/lib/watch/index.js#L42).

Tony133:
Resolved by #851, which removed the dotenv dependency from fastify-cli (also closing #559, the previous attempt to fix this same issue). With dotenv removed, the precedence issue described here no longer occurs.

PR #559, opened 2022-10-14
manual loading and parsing of dotenv in watch mode #558

tandibar:
I think this could be the fix for the issue #558 that I opened.

The problem with `dotenv.config().parsed` is that `dotenv.config()` modifies the `process.env`. So the fix uses `dotenv.parse()` directly so that `process.env is untouched.

All tests are green. I will provide an additional test in the next days.

#### Checklist

- [x] run `npm run test` ~and `npm run benchmark`~
- [ ] ~documentation is changed or added~

### DIFF
--- a/lib/watch/index.js
+++ b/lib/watch/index.js
@@ -3,6 +3,7 @@
 const path = require('path')
 const cp = require('child_process')
 const chalk = require('chalk')
+const { readFileSync, existsSync } = require('fs')
 const { arrayToRegExp, logWatchVerbose } = require('./utils')
 const { GRACEFUL_SHUT } = require('./constants.js')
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
Thanks for opening a PR! Can you please add a unit test?

mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

tandibar:
So I added a unit test and implemented the feature. The problem is, that it breaks another feature, see test `test/start.test.js should reload the env on restart when watching`. In the current implementation I think it is very hard to have both features. For me it would be more consistent if dotenv does not change its behavior in watch mode. But that would mean, that you have to restart your server if you change your `.env` even in watch mode.
What do you think?

mcollina:
> that would mean, that you have to restart your server if you change your .env even in watch mode.

Why? I _think_ it might be possible to have the env variables defined in process.env to have priority over the ones in .env.
Or viceversa. I think the surprising behavior is lack of consistency.

tandibar:
So the problem ist, that `dotenv` is already loaded before, so we cannot just check which env vars are already present, because all are present. So I decided to check if there is a difference between the `process.env` and the parsed dotenv. If yes, we should not overwrite the env var. If it is the same, it seems to be managed by dotenv file, so we should overwrite.
Tests are now all passing, so I think that is the solution.

PR #568, opened 2022-11-20
fix: server config deep merge

Eomm:
requires https://github.com/fastify/deepmerge/pull/13

when the developer has a custom server option and needs to customize it when running the tests,  it is impossible to do it easily because the test options overwrite all the one set by the CLI arguments


Uzlopak (approved):
LGTM

Eomm:
@fastify/plugins need another review 🙏🏽

mcollina (approved):
lgtm

Issue #579, opened 2023-01-10
isDocker is not working on Kubernetes v1.23


QuanTran91:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.1.1

### Node.js version

16

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

20.04

### Description

Plugin [is docker](https://github.com/sindresorhus/is-docker) is not work on EKS (Elastic Kubernetes Service) 1.23

### Steps to Reproduce

Run app in EKS as pod
Observe the log that it is running on 127.0.0.1

climba03003:
I would say it is a upstream issue. and you are allowed to pass the address by yourself.

sinedied:
I would say that isDocker works as expected since K8s doesn't implement a Docker engine, but it would be nice if the fastify-cli adds some detection logic to use 0.0.0.0 as default in K8s environments. I lost 2h yesterday because of this

mcollina:
Ah! That's the catch. Do you know how can we detect that we running on K8s?

sinedied:
Seems like you can check for an environment variable: https://stackoverflow.com/questions/36639062/how-do-i-tell-if-my-container-is-running-inside-a-kubernetes-cluster

sinedied:
Would you accept a PR for this?

climba03003:
> Would you accept a PR for this?

Yes, and I see more complete check on this package https://github.com/ntfs32/is-kubernetes/blob/main/index.js
The download is 1 / week, so I have a bit worry on the stability. But, the detection look solid to me.

sinedied:
Ok, I'll propose a PR then. Though I would be against adding an external dependency for something that is a 1-liner, as the environment variable is enough (according to the docs, it's always present: https://kubernetes.io/docs/tutorials/services/connect-applications-service/#environment-variables).

climba03003:
> Though I would be against adding an external dependency for something that is a 1-liner, as the environment variable is enough

I have no idea will this change in future but multiple method for detection doesn't hurt.

sinedied:
Given that any change to this would break the entire k8s ecosystem, it's very unlikely.
I'd rather thrust the official doc than than an unknown repo using a detection methods based on synchronous file reads, which could delay the startup a bit. 

I've sent the PR BTW.

climba03003:
[...]

PR #580, opened 2023-01-12
fix: use address 0.0.0.0 in Kubernetes env (#579)

sinedied:
#### Checklist

This PR adds a check so that it automatically use 0.0.0.0 when running inside a Kubernetes environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ You can pass the following options via CLI arguments. You can also use `--config
 | Path to logging configuration module to use                                                                                             | `-L`          | `--logging-module`  | `FASTIFY_LOGGING_MODULE` |
 | Start Fastify app in debug mode with nodejs inspector                                                                                   | `-d`          | `--debug`          | `FASTIFY_DEBUG`          |
 | Set the inspector port (default: 9320)                                                                                                  | `-I`          | `--debug-port`     | `FASTIFY_DEBUG_PORT`     |
-| Set the inspector host to listen on (default: loopback address or `0.0.0.0` inside Docker)                                              |               | `--debug-host`     | `FASTIFY_DEBUG_HOST`     |
+| Set the inspector host to listen on (default: loopback address or `0.0.0.0` inside Docker or Kubernetes)                                              |               | `--debug-host`     | `FASTIFY_DEBUG_HOST`     |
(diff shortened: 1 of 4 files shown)


climba03003 (commented):
I still in favor of multiple methods for the detection.

For example, different cloud platform may result in derivation of behavior which means that this environment variables are not always set.
https://stackoverflow.com/questions/66470062/aks-not-creating-kubernetes-service-host-and-kubernetes-service-port-environment

climba03003:
A proper implementation should be follow [`kubectl` check](https://kubernetes.io/docs/reference/kubectl/#in-cluster-authentication-and-namespace-overrides).

> By default kubectl will first determine if it is running within a pod, and thus in a cluster. It starts by checking for the KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables and the existence of a service account token file at /var/run/secrets/kubernetes.io/serviceaccount/token. If all three are found in-cluster authentication is assumed.

> To maintain backwards compatibility, if the POD_NAMESPACE environment variable is set during in-cluster authentication it will override the default namespace from the service account token. Any manifests or tools relying on namespace defaulting will be affected by this.

We are not authenticating for `in-cluster` but the method is good for environment detection.

mcollina (approved):
lgtm

sinedied:
I'll update the PR based on this

> A proper implementation should be follow [`kubectl` check](https://kubernetes.io/docs/reference/kubectl/#in-cluster-authentication-and-namespace-overrides).
> 
> > By default kubectl will first determine if it is running within a pod, and thus in a cluster. It starts by checking for the KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables and the existence of a service account token file at /var/run/secrets/kubernetes.io/serviceaccount/token. If all three are found in-cluster authentication is assumed.
> 
> > To maintain backwards compatibility, if the POD_NAMESPACE environment variable is set during in-cluster authentication it will override the default namespace from the service account token. Any manifests or tools relying on namespace defaulting will be affected by this.
> 
> We are not authenticating for `in-cluster` but the method is good for environment detection.

Good point, I'll update the PR with the additional check. I was looking for something similar in the docs, but couldn't find it 🙂

climba03003 (approved):

PR #589, opened 2023-02-21
fix: Correct wrong instructions for typescript ejecting

anyonecancode:
As written, the help text says that a user can run `eject-ts` to create a standalone `server.ts`. However, this is incorrect. Doing so simply creates a `server.js` -- all text following `eject` is ignored. For instance, one could run `eject-foo` and get no error and have a `server.js` generated, even though there is no `eject-foo` command. 

Instead, a user must pass in the `lang` flag: https://github.com/fastify/fastify-cli/blob/426e359f65e51a054fa89117ce729f34f88a6c37/eject.js#L25-L26

The easiest way to update this help test looks to be to use the `ts` rather than `typescript` value for `--lang`, so that it can easily fit into the existing help text format.

#### Checklist

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -2,7 +2,7 @@ Fastify command line interface available commands are:
 
   * start                 start a server
   * eject                 turns your application into a standalone executable with a server.js file being added.
-  * eject-ts              turns your application into a standalone executable with a server.ts file being added.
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

Issue #608, opened 2023-03-27
generate-swagger doesn't work with typescript code


smaant:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.15.0

### Node.js version

v19.7.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

Monterey (12.5.1)

### Description

I'm trying to generate swagger using fastify-cli for a Typescript project and getting an error `SyntaxError: Cannot use import statement outside a module`.

### Steps to Reproduce

1. Create a new Fastify project from template:
```
$ fastify generate --lang=ts test-project
```

2. Run `npm install`

4. Add `@fastify/swagger` - `npm install @fastify/swagger`

5. In the `src/app.ts` add:
```ts
+ import swagger from '@fastify/swagger';

// [skipped]

const app: FastifyPluginAsync<AppOptions> = async (
    fastify,
    opts
): Promise<void> => {
  // Place here your custom code!
+  fastify.register(swagger, {});

// [skipped remaining]
```

6. Try generating swagger:
```
$ fastify generate-swagger src/app.ts
```

7. Observe the error
```
import { join } from 'path';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1195:20)
    at Module._compile (node:internal/modules/cjs/loader:1239:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1329:10)
    at Module.load (node:internal/modules/cjs/loader:1133:32)
    at Module._load (node:internal/modules/cjs/loader:972:12)
    at Module.require (node:internal/modules/cjs/loader:1157:19)
    at require (node:internal/modules/helpers:119:18)
    at requireServerPluginFromPath (/Users/anton/.nvm/versions/node/v19.7.0/lib/node_modules/fastify-cli/util.js:79:20)
    at async runFastify (/Users/anton/.nvm/versions/node/v19.7.0/lib/node_modules/fastify-cli/generate-swagger.js:60:12)
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

yazaldefilimonepinto:
> Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

I would like try it...

[...]

PR #641, opened 2023-07-21
Feature/support generate swagge for typescript (#608)

yazaldefilimonepinto:
#### Checklist


yazaldefilimonepinto:
#608

util.js
```diff
@@ -55,33 +58,54 @@ async function getPackageType (cwd) {
...
 }
 
 async function requireServerPluginFromPath (modulePath) {
   const resolvedModulePath = path.resolve(process.cwd(), modulePath)
 
   if (!fs.existsSync(resolvedModulePath)) {
-    throw new Error(`${resolvedModulePath} doesn't exist within ${process.cwd()}`)
+    throw new Error(
+      `${resolvedModulePath} doesn't exist within ${process.cwd()}`
+    )
```
mcollina, line 76: why this change?

util.js
```diff
@@ -55,33 +58,54 @@ async function getPackageType (cwd) {
 function getScriptType (fname, packageType) {
   const modulePattern = /\.mjs$/i
   const commonjsPattern = /\.cjs$/i
-  return (modulePattern.test(fname) ? 'module' : commonjsPattern.test(fname) ? 'commonjs' : packageType) || 'commonjs'
+  return (
+    (modulePattern.test(fname)
+      ? 'module'
+      : commonjsPattern.test(fname)
+        ? 'commonjs'
+        : packageType) || 'commonjs'
```
mcollina, line 66: why this change?

util.js
```diff
@@ -4,8 +4,11 @@ const url = require('url')
 const semver = require('semver')
 const pkgUp = require('pkg-up')
 const resolveFrom = require('resolve-from')
-
-const moduleSupport = semver.satisfies(process.version, '>= 14 || >= 12.17.0 < 13.0.0')
+const tsNode = require('ts-node')
+const moduleSupport = semver.satisfies(
+  process.version,
+  '>= 14 || >= 12.17.0 < 13.0.0'
+)
```
mcollina, line 11: why this change?

mcollina (changes requested):
Thanks for sending a PR! I don't think we should add `ts-node` as a dependency :(, as it's too opinionated and subjected to TypeScript semver rules. 
Note that the [generator](https://github.com/fastify/fastify-cli/blob/master/generate.js#L48) does not use `ts-node` either. 

I would recommend you try a different approach:
1. load `tsconfig.json` and check the `outDir` folder
2. load the application from there
3. if `outDir` is not built, error telling the user to build the folder
4. adjust the typescript template to add a "swagger" definition

---

There seems to be quite a few linting issues that are spurious. Did you enable prettier format-on-save for all files in VSCode?

yazaldefilimonepinto:
> Thanks for sending a PR! I don't think we should add `ts-node` as a dependency :(, as it's too opinionated and subjected to TypeScript semver rules. Note that the [generator](https://github.com/fastify/fastify-cli/blob/master/generate.js#L48) does not use `ts-node` either.
> 
> I would recommend you try a different approach:
> 
> 1. load `tsconfig.json` and check the `outDir` folder
> 2. load the application from there
> 3. if `outDir` is not built, error telling the user to build the folder
> 4. adjust the typescript template to add a "swagger" script
> 
> There seems to be quite a few linting issues that are spurious. Did you enable prettier format-on-save for all files in VSCode?

Thanks for the feedback! I appreciate your guidance, and I've already implemented the suggested, regarding the linting issues, I've taken care of them. 
I'm glad to contribute to the project's improvement, I hope I can still bring great contributions!

mcollina:
Code LGTM, a test is missing.

yazaldefilimonepinto:
> 

Hello, sorry for make you await, but you can tell me, how test it, i think test is about confirm if genereta ts incluede swagger script

mcollina:
> Hello, sorry for make you await, but you can tell me, how test it, i think test is about confirm if genereta ts incluede swagger script

Yes.

I think you should add the required logic in https://github.com/fastify/fastify-cli/blob/master/test/generate-typescript.test.js.

I would recommend adding a test similar to https://github.com/fastify/fastify-cli/blob/a444cd85173034f7e2d5067e680f243c07492e31/test/generate-typescript.test.js#L103 that verifies that the file is generated for the TS app.

mcollina:
The test is still missing. You would want to add a new test that calls that new command that you added.

PR #652, opened 2023-08-10
fix: app template options

lqqyt2423:
Pass --options via CLI arguments in command have no effect


cescobaz:
I spent 1 hour to debug why it does not work!!
I applied your change to my project and it fixes the issue. Thanks!

climba03003 (approved):
Yes, the order is important.

andresayac (approved):
haha I hope it is approved hahaha it definitely gives confusion

PR #654, opened 2023-08-10
Fix/connot set trust proxy

yazaldefilimonepinto:
#### Checklist


yazaldefilimonepinto:
#654 619

mcollina (approved):
lgtm

args.js
```diff
@@ -23,13 +23,14 @@ const DEFAULT_ARGUMENTS = {
 
 module.exports = function parseArgs (args) {
   dotenv.config()
+  const TRUST_PROXY = process.env.TRUST_PROXY || ''
```
Eomm, line 26: We use the FASTIFY_ prefix everywhere, I would keep it

args.js
```diff
@@ -23,13 +23,14 @@ const DEFAULT_ARGUMENTS = {
...
   const commandLineArguments = argv(args, {
     configuration: {
       'populate--': true
     },
     number: ['port', 'inspect-port', 'body-limit', 'plugin-timeout', 'close-grace-delay'],
     string: ['log-level', 'address', 'socket', 'prefix', 'ignore-watch', 'logging-module', 'debug-host', 'lang', 'require', 'config', 'method'],
-    boolean: ['pretty-logs', 'options', 'watch', 'verbose-watch', 'debug', 'standardlint', 'common-prefix', 'include-hooks'],
+    boolean: ['pretty-logs', 'options', 'trustProxy', 'watch', 'verbose-watch', 'debug', 'standardlint', 'common-prefix', 'include-hooks'],
```
Eomm, line 33: it has multiple shapes.

fwis, string and boolean are the most used so I would support it or I would rename it to `FASTIFY_ENABLE_TRUST_PROXY`

args.js
```diff
@@ -65,7 +66,9 @@ module.exports = function parseArgs (args) {
 
   // Merge objects from lower to higher priority
   const parsedArgs = { ...DEFAULT_ARGUMENTS, ...configFileOptions, ...commandLineArguments }
-
+  const isNumber = isNaN(Number(TRUST_PROXY)) === false
+  const isTrue = TRUST_PROXY.toLowerCase() === 'true'
+  const trustProxy = isNumber ? Number(TRUST_PROXY) : isTrue
```
Eomm, line 71: this is very hard to read and it does too much logic IMHO

regardless this code, the `parsedArgs` will win if the user set a boolean.. I would hardly understand the behaviour here.

KIS approach:
- one env per boolean
- one env per trust proxy hosts
- one env per trust hop

the priority is the list i wrote when user set multiple env vars

Eomm (changes requested):

twalling:
@Eomm it seems this PR was closed, we use `fastify-cli` in production and need to set `trustProxy` now so I'd be happy to submit a PR based on the feedback it seems you had

Am I correct in reading this that for this more flexible `fastify` argument the preference would be 2-3 explicit arguments (named appropriately) for the various use cases and then just go through choosing the one to use based on the priority list you mentioned?

Issue #593, opened 2023-03-03
TypeScript AppOptions are not provided to app unless passed via CLI


evanshortiss:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.13.0

### Node.js version

18.12.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.6.3

### Description

The TypeScript template allows us to provide options using the following in *app.ts*, however the `opts` variable passed to the `app` object doesn't contain my hardcoded options, e.g `foo` in the sample code below. Options provided with the CLI **do** appear in the object. Sample code:

```ts
export type AppOptions = {
  foo: string
  // Place your custom options for app below here.
} & Partial<AutoloadPluginOptions>;


// Pass --options via CLI arguments in command to enable these options.
const options: AppOptions = {
  foo: 'bar'
}

const app: FastifyPluginAsync<AppOptions> = async (
    fastify,
    opts
): Promise<void> => {
  // Place here your custom code!
  console.log(opts.foo) // should print "bar"

  // Do not touch the following lines

  // This loads all plugins defined in plugins
  // those should be support plugins that are reused
  // through your application
  void fastify.register(AutoLoad, {
    dir: join(__dirname, 'plugins'),
    options: opts
  })

  // This loads all plugins defined in routes
  // define your routes in one of these
  void fastify.register(AutoLoad, {
    dir: join(__dirname, 'routes'),
    options: opts
  })
};
```

The example `SupportPluginOptions` in the TS template receives an object like so: `{ prefix: undefined }`, which doesn't align with the `SupportPluginOptions`.

I might be opening a can of worms here about how the options functionality is supposed to work? The reason I ran into this issue is because I wanted to supply environment variables to my plugins via the `options` object. The solution I came up with was to modify the code to load plugins like so:

```ts
const options: AppOptions = {
  foo: 'bar',
  env: {
    maxValue: process.env.MAX_VALUE
  }
}


// Then inside the "app" function I use object.assign
void fastify.register(AutoLoad, {
  dir: join(__dirname, 'plugins'),
  options: Object.assign({ options }, opts)
})
```

### Steps to Reproduce

1. Generate a TS project: `npx fastify-cli@5.7.1 generate --lang=ts autoload-options`
1. Define custom `AppOptions` with some defaults (see example above)
1. Run code and notice that options are missing

### Expected Behavior

[...]

PR #728, opened 2024-06-06
fix: ensure `options` are passed to app plugin with `--options` flag

dhensby:
The TS typings and docs strongly imply that the `options` object defined in the app.js file should make their way to the app plugin if the `--options` cli argument is present when running `fastify start app.js`. This fix does just that, providing the exported options object back to the app plugin when it is registered.

The CLI allows arguments to make their way to the app via additional args (coming after a `--`), eg: `fastify start app.js -l info -- --custom test` will populate the plugin `options` argument with a `custom: "test"` property. This change deliberately leaves precedence with the CLI args so that the exported `options` object just forms the default values which can then be overridden as required.

Potential fix to #593 

#### Checklist

- [x] run `npm run test` ~and `npm run benchmark`~

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ module.exports = async function (fastify, options) {
 For a list of available flags for `fastify start` see the help: `fastify help start`.
 
 If you want to use custom options for the server creation, just export an options object with your route and run the cli command with the `--options` flag.
+These options also get passed to your plugin via the `options` argument.
 
 ```js
 // plugin.js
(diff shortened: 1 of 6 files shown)


dhensby:
It does seem quite deliberate at the moment that these `options` don't make their way to the plugin and seem to not actually be `AppOptions` but instead `FastifyServerOptions`. However, the TS typings definitely set the expectation that the options would come be populated from them when using the `--options` flag.

see the TS template for the misleading typings: https://github.com/fastify/fastify-cli/blob/v6.1.1/templates/app-ts/src/app.ts

dhensby:
Fixed the linting failure, not sure why the tests failed on macos runner - I've tried adding `t.end()` to the end of the new tests incase that was needed, but otherwise can't see an obvious problem 🤔

dhensby:
Fixed the linting regression from adding `t.end();` 😓

jsumners:
attn: @fastify/typescript

mcollina (approved):
lgtm

Issue #168, opened 2019-04-19
error occurs when writting files inside application directory with 'fastify start --watch'


tranzwalle:
I writting a file under the appliction directory, like 'cache/filecache.json',
and start the application with command 'fastify start --log-level info --pretty-logs app.js'.
If the file was changed, an error will show in terminal, "[fastify-cli] process forced end"

of course, this error can be avoided by '--ignore-watch'.
But when I use nodemon, there's no such error, so I submit this

tranwalle:
<img width="386" alt="fastify-cli-err" src="https://user-images.githubusercontent.com/8072111/56414187-60dd4b80-62bc-11e9-8eb9-d1bc6c28c6fb.png">

mcollina:
Ouch! Would you like to send a PR to fix this?

tranwalle:
OK, it maybe take times, I haven't read the source code.
As a new user of fastify, I think 'fastify-cli' shoud be simple and can be integrated with other awesome tools.  How about let PM2 and nodemon to take over the feature of '--watch'

zuck:
Same here. Any hint about what could generate this issue?

mcollina:
No clue, it would be good to have an example, even better a PR with a failing test.

zuck:
It's definitely an issue with big files changing inside the folder the CLI is watching. In my case was a SQLite DB file.

The workaround is the same mentioned above, just use the `--ignore-watch FILE_NAME` option to _fix_ it.

whoisjuan:
@mcollina This went stale, but definitely a still problem. Writing a file inside the application directory crashes the CLI. This also makes you lose visibility of any logs that happen between the process ending and the server restarting to listen.

mariusa:
+1 still an issue, shouldn't be closed

mariusa:
I'm curious if this is a bug. I came here because I didn't know what caused it. After seeing the cause, `FASTIFY_IGNORE_WATCH` was the proper solution. If this were present in the error message ` process forced end`, it would have saved me time.

I believe `--watch` does what it asked, to watch for changes and restart fastify. Changing a file and complaining it restarts shouldn't be an issue, but the error message should mention the cause & solution.

Eomm:
> I'm curious if this is a bug

Me too 😄

We don't have a failing error test or something that we can just execute to reproduce the error. I'm using `fastify-cli` and I did not hit this error, so I can't invest too much time to replicate it.

[...]

PR #765, opened 2024-09-27
feat: pass forceCloseConnections option from cli

LucaRainone:
Enabling the proxy for the  [forceCloseConnections](https://fastify.dev/docs/latest/Reference/Server/#forcecloseconnections) option. When enabled, this option resolves [this issue](https://github.com/fastify/fastify-cli/issues/168). Explained in detail also inside [this other issue](https://github.com/fastify/fastify/issues/5713)

Please note that the original option on fastify could be both `string` and `boolean`. In fastify-cli I didn't find a smart way to manage it (it seems that's possible defining only one type in a structural way). I choose to use it as boolean/flag, leaving the only allowed string value (`idle`) the default option. 
If there is a better standard solution, please point me in the right direction :) 

#### Checklist


README.md
```diff
@@ -149,28 +149,29 @@ If your `package.json` does not have `"type": "module"`, use `.mjs` for the exte
...
+| Set the plugin timeout                                                                                                                                                           | `-T`          | `--plugin-timeout`          | `FASTIFY_PLUGIN_TIMEOUT`          |
+| Set forceCloseConnections option on fastify instance                                                                                                                             | `-f`          | `--force-close-connections` | `FASTIFY_FORCE_CLOSE_CONNECTIONS` |
+| Defines the maximum payload, in bytes,<br>that the server is allowed to accept                                                                                                   |               | `--body-limit`              | `FASTIFY_BODY_LIMIT`              |
+| Set the maximum ms delay before forcefully closing pending requests after receiving SIGTERM or SIGINT signals; and uncaughtException or unhandledRejection errors (default: 500) | `-g`          | `--close-grace-delay`       | `FASTIFY_CLOSE_GRACE_DELAY`       |
```
mcollina, line 174: Why did you reformat the whole table? Can you only make the change needed?

LucaRainone: Ops, I believed was some linter. Instead was my IDE. Restored the untouched lines

mcollina (approved):
lgtm

Eomm (commented):
TBH I would set this config to `true` by default because it may cause data loss in production

- As is: an error logging message
- With default `true`: a request that is dropped in the middle of processing

As a user, I would be more pissed of about the 2nd default scenario.

Can't we write in the log:

```
process forced end. There were 2 open connections. Checkout the `forceCloseConnections` option
```

LucaRainone:
> TBH I would set this config to `true` by default because it may cause data loss in production
> 

I don't get it.
If `forceCloseConnections` is true then it closes all pending connections regardless their status, causing possibly data loss. That's why it's not the default value.

If `forceCloseConnections` is not set in `fastify-cli` then it's not passed at all to `fastify` instance, keeping the default value of `fastify` option (that's `idle` or false depending by the server, so it tries a graceful shutdown limiting the data loss).

We can write additional infos on message, but this message could also shown in production during restart with "real" chrome connections opened: may be a false hint?

Issue #781, opened 2024-12-09
TypeScript type augmentations not visible in tests with ESM modules


MatthewDahl:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.0.0

### Plugin version

7.0.1

### Node.js version

20.17.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

macOS 15.1.1, build: 24B91

### Description

# TypeScript type augmentations not visible in tests with ESM modules

## Problem
When using ESM modules (`--esm`) with TypeScript (`--lang=ts`), plugin type augmentations aren't visible during test execution. This specifically affects projects where plugins depend on other plugins' type augmentations, causing runtime errors in tests despite TypeScript compilation succeeding.

## Reproduction Case
1. Create a new project:
```bash
fastify generate my-app --esm --lang=ts
```

2. Add a base plugin with type augmentations:
```typescript
// src/plugins/config.ts
import fp from 'fastify-plugin'

export interface Config {
  value: string
}

export default fp(async (fastify) => {
  fastify.decorate('config', { value: 'test' })
}, { name: 'config' })

declare module 'fastify' {
  interface FastifyInstance {
    config: Config
  }
}
```

3. Add a dependent plugin that uses the config type:
```typescript
// src/plugins/feature.ts
import fp from 'fastify-plugin'

export default fp(async (fastify) => {
  const configValue = fastify.config.value
  fastify.decorate('feature', { configValue })
}, {
  name: 'feature',
  dependencies: ['config']
})

declare module 'fastify' {
  interface FastifyInstance {
    feature: {
      configValue: string
    }
  }
}
```

4. Add a test that uses the plugin types:
```typescript
// test/plugins/feature.test.ts
import { test } from 'node:test'
import * as assert from 'node:assert'
import { build } from '../helper.js'

test('feature plugin type augmentations', async (t) => {
  const app = await build(t)

  // This line should cause TypeScript errors without proper type augmentations
  const value: string = app.config.value
  assert.equal(value, 'test')

  // This should also cause TypeScript errors
  const configValue: string = app.feature.configValue
  assert.equal(configValue, 'test')
})
```

5. Run the tests:
```bash
npm test
```

The test fails with:
```
✖ feature plugin type augmentations
  TypeError [Error]: Cannot read properties of undefined (reading 'value')
[...]

Issue #628, opened 2023-05-24
generated build function for test setup doesn't load/autoload plugins


cherifGsoul:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

4.x

### Stopped working in version

5.7.1

### Node.js version

v18.16.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 22.04.2 LTS

### 💥 Regression Report

Generated app, in the test build function, I believe `const app = await helper.build(argv, await config())` doesn't load/autoload plugins, Am I missing something?
The issue is about the following line https://github.com/fastify/fastify-cli/blob/f111b25b96d9f29d42c9a8f5ced4108f56a073d0/templates/app-ts/test/helper.ts#L24
app.register app.ready work perfectly before.

In my case, I needed this in the test setup to use the ORM fastify plugin I'm using to generate db schema and remove in teardown.

### Steps to Reproduce

- Generate a new app with fastify-cli
- Try to access a plugin function inside the build function itself eg: `app.someSupport()`
```js
async function build (t: Test) {
  // you can set all the options supported by the fastify CLI command
  const argv = [AppPath]

  // fastify-plugin ensures that all decorators
  // are exposed for testing purposes, this is
  // different from the production setup
  const app = await helper.build(argv, await config())
  app.someSupport()

  // Tear down our app after we are done
  t.teardown(() => void app.close())
  return app
}
```
- An error should be thrown

### Expected Behavior

The plugin function results should be returned.

Tony133:
This problem is resolved in this PR here: https://github.com/fastify/fastify-cli/pull/784

Issue #622, opened 2023-05-08
Can't access plugin inside tests


thales-maciel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I've generated the project using the cli and then I modified my app.ts file so that it looks like this:
```
import { join } from 'path';
import AutoLoad, {AutoloadPluginOptions} from '@fastify/autoload';
import { FastifyInstance, FastifyPluginAsync } from 'fastify';
import { fastifyPostgres } from '@fastify/postgres';
import fastifySwagger from '@fastify/swagger';
import fastifySwaggerUi from '@fastify/swagger-ui';
import sensible from '@fastify/sensible';
import cors from '@fastify/cors'

export type AppOptions = {} & Partial<AutoloadPluginOptions>;


// Pass --options via CLI arguments in command to enable these options.
const options: AppOptions = {}

const app: FastifyPluginAsync<AppOptions> = async (fastify: FastifyInstance, opts: any): Promise<void> => {
  fastify.register(sensible)

  fastify.register(fastifyPostgres, {
    connectionString: `postgres://${process.env.DB_USER}:${process.env.DB_PASSWORD}@${process.env.DB_HOST}/${process.env.DB_DATABASE}`
  })

  fastify.register(fastifySwagger, { hideUntagged: true })
  fastify.register(fastifySwaggerUi, {
    routePrefix: '/docs',
    uiConfig: { docExpansion: 'list' }
  })

  await fastify.register(cors, {})

  void fastify.register(AutoLoad, {
    dir: join(__dirname, 'routes'),
    options: opts
  })

}

export default app;
export { app, options }
```

The problem is that when i try to use the pg plugin inside a test I get an error `Cannot read properties of undefined (reading 'query')`, which makes me believe that the plugin hasn't been registered, but I don't know how to do it.

fech-dev:
I have notice that the `build()` function does not wrap the app plugin (exported in src/app.ts) with `fp()` function.
I've tried to apply fp function and I can access all decorators defined. 
I don't know if this is intentional or not....

Hornwitser:
[...]

PR #784, opened 2024-12-15
refactor: generate test build function configured with skipOverride

jean-michelet:
I think that users expect to have access decorators defined globally in their tests.
Currently we generate the test application without wrapping it in `fastify-plugin` which  causes confusions (c.f. [here](https://github.com/fastify/fastify-cli/issues/622), [here](https://github.com/fastify/fastify-cli/issues/628) and [here](https://github.com/fastify/fastify-cli/issues/781)).

### DIFF
--- a/templates/app-ts-esm/test/helper.ts
+++ b/templates/app-ts-esm/test/helper.ts
@@ -14,7 +14,7 @@ const AppPath = path.join(__dirname, '..', 'src', 'app.ts')
 
 // Fill in this config with all the configurations
 // needed for testing the application
-async function config () {
+function config () {
   return {
     skipOverride: true // Register our application with fastify-plugin
   }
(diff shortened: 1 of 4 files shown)


templates/app-ts-esm/test/helper.ts
```diff
@@ -15,7 +15,9 @@ const AppPath = path.join(__dirname, '..', 'src', 'app.ts')
 // Fill in this config with all the configurations
 // needed for testing the application
 async function config () {
```
jean-michelet, line 17: Btw, why `config` function return promises in ts templates?
Should I replace by a sync function?

mcollina (approved):
lgtm

templates/app-ts-esm/test/helper.ts
mcollina: I think because that's the signature.

mcollina (approved):
lgtm

Issue #756, opened 2024-08-28
Typescript Fastify app generated with cli has error when running `npm run test`


ascodeasice:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.28.1

### Node.js version

18.20.4

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Linux Mint  21.3

### Description

After running the following commands:

```bash
fastify generate fastify-test --lang=ts
cd fastify-test
npm install
```

I encountered an error when executing the `npm test` command:

> Could not find '/home/leo/test/fastify-test/test/**/*.ts'

To fix this issue, I modified the `test` script in `package.json` by removing the double quotes around the file path:

```diff
- "test": "npm run build:ts && tsc -p test/tsconfig.json && c8 node --test -r ts-node/register \"test/**/*.ts\"",
+ "test": "npm run build:ts && tsc -p test/tsconfig.json && c8 node --test -r ts-node/register test/**/*.ts",
```

After making this change, the error was resolved, and the tests ran without issues.

Upon further investigation, I discovered PR https://github.com/fastify/fastify-cli/pull/708 and learned that for Node.js versions prior to 21, commands with double quotes around file paths don't work correctly, and such versions (without double quotes) can only identify files at the first directory level. However, in Node.js versions 21 and later, commands work both with and without double quotes. Moreover, using double quotes allows files at deeper directory levels to be recognized.

**TL;DR:** To run tests reliably, use Node.js version 21 or newer.

mcollina:
Sorry for the long wait... this sat in my inbox since August.

Is this still valid? Is there something we should do about it? It seems a npm limitation more than anything else.. and something it's already solved.

vogloblinsky:
@mcollina still valid, just got it today CLI

PR #802, opened 2025-03-10
fix: correctly handle glob patterns in test script

skrzepij:
**PR Summary:** 
Fix Test Pattern Glob Matching in fastify-cli

**Issue**
The current test script in fastify-cli has a problem with glob pattern matching. When using the pattern `\"test/**/*.ts\"` in the `npm test` script, the pattern is not properly interpreted as a glob expression, resulting in no test files being found and executed.  

**Fix**
This PR adjusts how test file patterns are processed to ensure proper glob pattern matching. By modifying the pattern formatting in the test script, we ensure that test discovery works correctly across different operating systems and environments.

**Why This Matters**

This fix improves the developer experience by:  
- Ensuring tests are properly discovered and executed without manual workarounds
- Providing consistent behavior across different operating systems
- Saving developers time by avoiding confusing test execution failures
- Maintaining the expected behavior of standard glob patterns in npm scripts
- The change is minimal but significantly improves the reliability of the test execution process.

#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -47,7 +47,7 @@ const typescriptTemplate = {
   dir: 'app-ts',
   main: 'app.ts',
   scripts: {
-    test: 'npm run build:ts && tsc -p test/tsconfig.json && c8 node --test -r ts-node/register "test/**/*.ts"',
+    test: 'npm run build:ts && tsc -p test/tsconfig.json && c8 node --test -r ts-node/register test/**/*.ts',
     start: 'npm run build:ts && fastify start -l info dist/app.js',
     'build:ts': 'tsc',
     'watch:ts': 'tsc -w',


skrzepij:
After further investigation, I've discovered issue #756 which addresses the same problem I encountered.

It appears that the solution is more complex than simply removing quotes, as it depends on Node.js version:
- **For Node.js < 21**: Removing quotes helps resolve the immediate issue
- **For Node.js ≥ 21**: Keeping quotes is actually preferred as it enables finding test files in nested directories

Since this is a version-dependent behavior I agree that the best approach is to keep the current configuration with quotes intact and recommend Node.js 21+ for full functionality.

I'll be closing this PR. For my own project where I'm using Node.js < 21, I'll manually remove the quotes in my package.json as a temporary workaround until I can upgrade to Node.js 21+.

I'd be happy to contribute a documentation update instead if that would be helpful to clarify this behavior for other users.

PR #883, opened 2026-04-23
fix: clean dist/ before build:ts in TypeScript template

### LINKED ISSUE #878: Add clean step to generated build:ts script to remove stale compiled routes
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

When using `fastify generate` with `--lang=ts`, the generated `build:ts` script does not clean the `dist/` directory before compiling.

This means that if you delete a route file from `routes/` or a plugin file from `plugins/`, the compiled `.js` file remains in `dist/` and continues to be served by `@fastify/autoload` — even though the source file no longer exists.

### Suggested Fix

Adding a dedicated `clean` script and using it in `build:ts` would prevent stale compiled files from being served:
```json
"scripts": {
  "clean": "node -e \"require('fs').rmSync('dist', {recursive:true,force:true})\"",
  "build:ts": "npm run clean && tsc",
}
```

This would make the generated template behave more intuitively out of the box, particularly for developers new to the framework who may not be aware of how `tsc` handles its output directory.

### Motivation

The `fastify generate` command is designed to get developers up and running quickly. The generated template should behave intuitively out of the box, especially for developers new to the framework.

Without a clean step, deleting a file from `routes/` or `plugins/` gives the misleading impression that the change has had no effect — the endpoint or plugin continues to work as if the file still exists. This can cause significant confusion during development and is a non-obvious debugging problem for those unfamiliar with how `tsc` manages its output directory.

Many popular CLI scaffolding tools include a clean step in their generated build scripts for exactly this reason.

### Example

After running `fastify generate myapp --lang=ts`, the generated `package.json` would include:
```json
"scripts": {
  "clean": "node -e \"require('fs').rmSync('dist', {recursive:true,force:true})\"",
  "build:ts": "npm run clean && tsc",
}
```

This ensures that when a developer deletes a route or plugin file, it is no longer present in `dist/` after the next build, and will not be served by `@fastify/autoload`.

winnrz:
## What
Adds a `clean` script to the generated TypeScript template and runs it before `tsc` in `build:ts`, so stale compiled files are removed on every build.

## Why
Closes #878. Without this, deleting a file from `routes/` or `plugins/` leaves a stale `.js` in `dist/`, which `@fastify/autoload` continues to serve — giving the misleading impression that the deletion had no effect. This is a non-obvious debugging problem for developers new to Fastify.

## Changes
- Added `clean` script using inline `node -e` (no new dependency, cross-platform)
- Updated `build:ts` to run `clean` first
- Added a `logInstructions` entry for `clean` so it's discoverable
- Fixed a pre-existing typo in `logInstructions`: `npm build:ts` → `npm run build:ts`

The ESM TypeScript template inherits from `typescriptTemplate` via spread in `cli()`, so both `--lang=ts` and `--lang=ts --esm` pick up the change.

## Verification

**Stale file removal on a generated project:**

```bash
$ mkdir -p dist && echo "stale" > dist/stale.js
$ ls dist/
stale.js
$ npm run clean
$ ls dist/ 2>/dev/null || echo "dist/ removed"
dist/ removed
```

**Tests pass locally:**
- `npm test` passes end-to-end
- `npm run unit:ts-cjs` ✓
- `npm run unit:ts-esm` ✓

On Node 20 the template test suites are auto-skipped by the repo's `should-skip-test-suites.js`, but I ran `unit:ts-cjs` and `unit:ts-esm` directly and both pass.

## Note
`npm run build:ts` currently fails on a freshly generated project due to an unrelated `TS5011` rootDir error, which reproduces on `main` without this change. This PR does not affect that behavior — the `clean` step completes successfully before `tsc` is invoked.

## Alternatives considered
- `rimraf`: adds a dependency for what's essentially a one-liner
- `tsc --build --clean`: requires incremental/project-reference builds, which the current template doesn't use

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

> Note on "tests and/or benchmarks are included": no new tests added because the change is to the generated `scripts` object, which isn't asserted on in the existing test suite. The existing TS template tests (`unit:ts-cjs` and `unit:ts-esm`) pass. Happy to add coverage if reviewers prefer.


winnrz:
Force-pushed to fix the lint failure as the previous commit had unintentional reformatting.

ilteoood (approved):
LGTM

ilteoood (approved):
LGTM

Eomm (approved):

Issue #835, opened 2025-07-19
Generated Fastify ESM/TS app has broken `npm test` command (FASTIFY_AUTOLOAD_TYPESCRIPT)


Ragzouken:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

7.4.0

### Node.js version

24.4.0

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

11

### Description

Setup
```
fastify generate fastify-test --lang=ts --esm
cd fastify-test
npm install
```

Attempt
```
npm run test
```

Outcome
```
> fastify-test@1.0.0 test
> npm run build:ts && tsc -p test/tsconfig.json && FASTIFY_AUTOLOAD_TYPESCRIPT=1 node --test --experimental-test-coverage --loader ts-node/esm test/**/*.ts


> fastify-test@1.0.0 build:ts
> tsc

'FASTIFY_AUTOLOAD_TYPESCRIPT' is not recognized as an internal or external command, 
operable program or batch file.
```

### Link to code that reproduces the bug

see above

### Expected Behavior

`npm run test` should successfully run as generated.

chutientrong:
This error occurs because you're running a command with Linux/macOS syntax in the Windows Command Prompt (cmd).

The best and most common way to fix this is to use the `cross-env` library and update your test script by prefixing the environment variable with `cross-env`, like this:

`cross-env FASTIFY_AUTOLOAD_TYPESCRIPT=1`

riyap03:
Hi  @Ragzouken ✨,
I reproduced this on Windows and confirmed the FASTIFY_AUTOLOAD_TYPESCRIPT env var fails in cmd.
I tested with cross-env and it works cross-platform (Windows/Linux). Would you like me to open a PR updating the generated template’s test script to use cross-env?
– Riya 🌷

PR #895, opened 2026-06-12
fix: generate a cross-platform test command for TS ESM apps

TibJnt:
## Summary
- use `cross-env` in the generated TypeScript ESM `npm test` script so the template works on Windows
- add generator assertions that the template includes `cross-env` and still preserves the expected test glob

## Problem
Issue #835 reports that generated `--lang=ts --esm` apps emit `FASTIFY_AUTOLOAD_TYPESCRIPT=1 node ...` in `package.json`, which breaks on Windows because inline environment variable assignment is shell-specific.

## Validation
- `node suite-runner.js "test/generate-typescript.test.js"`
- `node suite-runner.js "test/generate-typescript-esm.test.js"`

## Notes
- The branch was pushed via the existing SSH fork remote.
- The repo-level `npm test` command currently hits an unrelated Node 26 / `yargs` ESM failure in the existing suite.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -64,6 +64,7 @@ const typescriptTemplate = {
   devDependencies: {
     '@types/node': cliPkg.devDependencies['@types/node'],
     c8: cliPkg.devDependencies.c8,
+    'cross-env': cliPkg.devDependencies['cross-env'],
     'ts-node': cliPkg.devDependencies['ts-node'],
     concurrently: cliPkg.devDependencies.concurrently,
     'fastify-tsconfig': cliPkg.devDependencies['fastify-tsconfig'],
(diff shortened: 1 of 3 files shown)


ilteoood (changes requested):
I think it would be better to avoid using cross-env, considering that has been marked as "finished" almost [6 years](https://github.com/kentcdodds/cross-env/issues/257) ago and is now marked as read-only

Tony133:
Thanks for the contribution! I agree with @ilteoood’s comment above (https://github.com/fastify/fastify-cli/pull/895#pullrequestreview-5042851194), so I don’t think we should move forward with this approach.

I’m going to close this PR.

PR #898, opened 2026-06-22
chore: replace http with https in urls

Fdawgs:
Upgraded all http urls to https where possible. Links have been tested and verified


Tony133 (approved):
lgtm

PR #909, opened 2026-08-06
Properly await --import modules

segevfiner:
#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/start.js
+++ b/start.js
@@ -84,7 +84,7 @@ async function preloadESModules (opts) {
   if (typeof opts.import === 'string') {
     opts.import = [opts.import]
   }
-  opts.import.forEach(async (m) => {
+  await Promise.all(opts.import.map(async (m) => {
     if (m) {
       /* This check ensures we ignore `-i ""`, trailing `-i`, or
        * other silly things the user might (inadvertently) be doing.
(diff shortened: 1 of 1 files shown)


kilisamemarisaaa (commented):
I confirmed that awaiting the preload operation fixes the current early-return behavior, but `Promise.all(opts.import.map(...))` starts every import concurrently and therefore loses preload order.

That differs from both repeated Node `--import` flags and this CLI's synchronous `-r` path. I reproduced it on Node 24.12.0 with two data-URL modules: the first waits in top-level await and then sets a global; the second reads it.

```text
node --import first --import second
native-second-sees:first

Promise.all([first, second].map((m) => import(m)))
parallel-second-sees:undefined
parallel-after:first
```

Order matters for preloads such as registering a polyfill/loader before a dependent module runs. Could these be awaited sequentially instead, for example with `for...of` around the existing `await requireESModule(m)`? The existing multiple-ES-module test could also make module 1 use top-level await and assert that module 2 observes its side effect; its current final-state assertions pass under both sequential and concurrent loading.

segevfiner:
@mcollina What do you think? I think he is right.

PR #775, opened 2024-12-01
feat: Support ESM logging module

segevfiner:
Fixes #760 

~~I also want to add support for taking the logger from the main plugin module instead, in the same way you can specify an options object as a named export on it when giving the `--options` flag. But I haven't decided on how to name the CLI flag and if it should have a short single letter alias. Any suggestions?~~ &ndash; This already exists as `options.logger` with `--options` enabled. But there is an issue with `deepmerge` and passing an already created `pino.Logger`.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

Issue #574, opened 2022-12-13
Pino object logger option gets cloned after #568


stefanosala:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.10.2

### Plugin version

5.7.0

### Node.js version

18.12.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.0.1

### Description

After #568 was merged, the new behavior of `deepmerge` clone the `options` object and tries to clone any instance of Pino we pass as option, modifying it into a simple object so that all the Pino settings are lost.

### Steps to Reproduce

Given the following fastify startup module:

```javascript
import pino from 'pino';

export default async function (fastify, options) {
  fastify.get('/', async (req, reply) => ({ hello: 'world' }));
}

const logger = pino({
  transport: {
    target: 'pino-pretty',
  },
});

export const options = {
  logger,
};
```

When using fastify-cli like this:

    node_modules/.bin/fastify start --options a.js

I get this output:

    {"level":30,"time":1670931640310,"pid":68940,"hostname":"Stefanos-MacBook-Air.local","msg":"Server listening at http://[::1]:3000"}
    {"level":30,"time":1670931640311,"pid":68940,"hostname":"Stefanos-MacBook-Air.local","msg":"Server listening at http://127.0.0.1:3000"}

### Expected Behavior

The expected output would be something like this instead:

    [12:41:03.421] INFO (69034): Server listening at http://[::1]:3000
    [12:41:03.422] INFO (69034): Server listening at http://127.0.0.1:3000

I would tend to think that `deepmerge` should be modified in order to clone by reference any type of object that is not a plain JS object, but I'd appreciate any feedback in order to try and provide a patch for this.

Eomm:
Thanks for reporting

Let me check what I broke

jordanebelanger:
> Thanks for reporting
> 
> Let me check what I broke

Would be great if this can get some attention, this was really hard to debug and a huge head scratcher while I tried to figure out why my logger was no longer working as expected in production.

Tom-Brouwer:
> Thanks for reporting
> 
> Let me check what I broke

Great, thanks for your response.

Like @jordanebelanger I have a custom logger defined in my options in app.js, and no logs were being generated anymore after an update of fastify-cli.

[...]

PR #911, opened 2026-08-28
fix: pass logger instances via loggerInstance (fastify v5 rejects instances in `logger`)

kilisamemarisaaa:
## Summary

fastify v5 only accepts a configuration object in the `logger` option and throws `FST_ERR_LOG_INVALID_LOGGER_CONFIG` when given an instance — instances must go through `loggerInstance`. fastify-cli converges user-provided logger instances — from `--logging-module` (#775), `--options` files (the setup described in #574), or server options — into `options.logger`, so passing a real Pino instance crashes `fastify start` at startup:

```
FastifyError [Error]: logger options only accepts a configuration object.
    code: 'FST_ERR_LOG_INVALID_LOGGER_CONFIG'
```

This was invisible to the suite because the only logger tests use a config-object fixture (`test/data/custom-logger.js` = `{ name, customLevels: {...} }`); the instance path had zero coverage.

**Fix**: after all option merges converge in `start.js`, route any logger value exposing `child()` to `loggerInstance` (4 lines). Config-object loggers (including the existing fixtures and the `{ level }` default) are untouched.

**Regression test**: `-L` with a fixture exporting a real Pino instance; asserts the logger level survives into `fastify.log`.

## Verification

- `npm run lint` clean
- Current head is one commit based directly on `main`; it does not include #910
- Important CI caveat: `suite-runner.js` on current `main` still calls glob's removed callback API, so the green JS-suite jobs exit without running the tests. #910 fixes that runner; after it lands, this branch must be updated and the logger regression test revalidated before merge.

### DIFF
--- a/start.js
+++ b/start.js
@@ -171,6 +171,13 @@ async function runFastify (args, additionalOptions, serverOptions, serverModule)
     options.trustProxy = opts.trustProxy
   }
 
+  // fastify v5 only accepts a configuration object in `logger`;
+  // logger instances must be passed via `loggerInstance`
+  if (options.logger && typeof options.logger.child === 'function') {
+    options.loggerInstance = options.logger
+    delete options.logger
+  }
+
   const fastify = Fastify(options)
 
   if (opts.prefix) {
--- a/test/data/custom-logger-instance.js
+++ b/test/data/custom-logger-instance.js
@@ -1,0 +1,5 @@
+'use strict'
+
+const pino = require('pino')
+
+module.exports = pino({ level: 'warn' })
--- a/test/start.test.js
+++ b/test/start.test.js
@@ -981,6 +981,17 @@ test('should support custom logger configuration in ESM', async t => {
   t.pass('server closed')
 })
 
+test('should support a logger instance from a logging module', async t => {
+  t.plan(2)
+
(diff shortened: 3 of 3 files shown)


Tony133 (changes requested):
There are the same commits as in this PR: https://github.com/fastify/fastify-cli/pull/910 Can you create a clean PR with the changes described? Or do a rebase?

kilisamemarisaaa:
Thanks for the review @Tony133 — you're right, the branch was cut from the #910
branch and carried its commits. I've rebuilt the branch on top of the latest
upstream/main so it now contains only the logger-instance change (single
commit); CI is rerunning on the clean history.

kilisamemarisaaa:
Quick follow-up: CI on the clean single-commit history is fully green (17 checks — 14 passed, 3 skipped, 0 failures). Ready for re-review whenever you have a moment, @Tony133 — thanks again for catching the stray commits.

kilisamemarisaaa:
Correction: the current head (53f6329) is already a clean single commit based directly on main (ea960d3); it does not contain #910. The code diff is isolated, but the green CI does not yet validate the new JS regression test: suite-runner.js on main still uses glob's removed callback API, so the JS suites exit without running tests (fixed in #910). I'll keep this branch clean and revalidate after #910 lands.

PR #789, opened 2025-01-12
Update help-me to remove dependency on glob/inflight

mcollina:
Fixes https://github.com/fastify/help/issues/1043

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #913, opened 2026-09-08
fix: restore yargs-parser compatible argument parsing

### LINKED ISSUE #912: Regression in CLI argument parsing after yargs-parser migration
The migration from yargs-parser to the custom node:util.parseArgs wrapper regresses supported CLI argument handling.

Reproduction:
```sh
node --test test/generate-swagger.test.js
```

Observed:
- `generate-swagger --yaml=true <plugin>` fails with `Unknown option '--yaml'` before command-specific parsing runs.
- Plugin options such as `--hello world` are parsed as `{ hello: true }` instead of preserving `world`.
- Explicit boolean values such as `--watch=false` are treated as `true`.

Expected:
The existing yargs-parser behavior should remain compatible: `--yaml` should be accepted by generate-swagger, plugin option values should be preserved, and explicit false boolean values should remain false.

Failing tests include `test/generate-swagger.test.js` and `test/helper.test.js`. The regression appears to have been introduced by commit `f2738f5`.

Tony133:
Proposal:

The migration to `util.parseArgs` (f2738f5) dropped several behaviors of yargs-parser:

- unknown options threw `Unknown option` — the parser is now non-strict by default (`strict` only when explicitly `true`), like yargs-parser. `generate-swagger`, `eject`, etc. pass through the main `args.js` parser with options it does not know (`--yaml=true`).
- unknown options did not consume their value (`--hello world` → `hello: true`). Now `hello: 'world'`; a value starting with `-` is still treated as a flag.
- booleans could not be set to `false`: `--watch=false`, `--watch false`, `--watch 0`, `-w false` are now honored. `util.parseArgs` cannot express `--bool=false`, so the normalizer collects them and applies them after parsing.
- everything after `--` was being re-normalized; it is now passed through untouched, so `'--'` and plugin options match the input.
- repeated string options (`-r a -r b`, `--import`) kept only the last value; `start.js` then crashed with `opts.require.forEach is not a function`. They are collected into arrays again from `tokens`.
- a string argv is accepted again (with a quote-aware tokenizer); `helper.build('./plugin.js -- --hello world')` relied on it.
- `test/args.test.js` restores `hello: 'world'` (the migration commit had changed the expectation to `hello: true`) and adds tests for each case above.

Test runner:

`suite-runner.js` never ran a single test: `glob` was bumped to v13 (#789) and the callback API no longer exists, so `glob(pattern, cb)` resolved a promise nobody awaited and the process exited 0. That is why CI stayed green through the regressions above. The runner now awaits the promise, exits 1 when no file matches, and ignores `node_modules` and the `test/workdir*` directories left behind by the generate tests.

With tests actually running, two more environment issues showed up:

- `pkg-up@5`, `is-docker@4` and `chalk@6` (bumped in #896) are ESM-only. They are loaded through `require(esm)` (`.default` / named export); the `is-docker` mocks in `test/start.test.js` are adjusted accordingly. Note: this requires Node ≥ 22.12 — if Node 20 must still be supported, these three dependencies need to be pinned to their CJS versions instead.
- TypeScript 6 no longer includes every `@types/*` automatically: `types: ["node"]` is added to the two test tsconfigs.
- `unit:ts-esm` fails on Node 24 with `--loader ts-node/esm` when `run()` spawns the test files; it now uses `module.register()` via `--import ./test/configs/register-ts-esm.mjs`, which also removes the `ExperimentalWarning`.

Also: `generate-plugin.js` no longer writes an empty `tstyche: {}` when the template has none.

Note:

- I'm also preparing a separate PR to ensure 100% tests coverage 🔥
- Supersedes #914: the `concurrency: 1` runner setting and the version-tolerant `is-docker` import come from there (co-authored with @mcollina).

Closes #912

### DIFF
--- a/generate-plugin.js
+++ b/generate-plugin.js
@@ -6,7 +6,7 @@ const {
 } = require('node:fs').promises
 const { existsSync } = require('node:fs')
 const path = require('node:path')
-const chalk = require('chalk')
+const chalk = require('chalk').default
 const generify = require('generify')
 const parseArgs = require('./lib/parse-args')
 const cliPkg = require('./package')
@@ -69,7 +69,9 @@ async function generate (dir, template) {
   pkg.scripts = Object.assign(pkg.scripts || {}, template.scripts)
   pkg.dependencies = Object.assign(pkg.dependencies || {}, template.dependencies)
   pkg.devDependencies = Object.assign(pkg.devDependencies || {}, template.devDependencies)
-  pkg.tstyche = Object.assign(pkg.tstyche || {}, template.tstyche)
+  if (template.tstyche) {
+    pkg.tstyche = Object.assign(pkg.tstyche || {}, template.tstyche)
+  }
 
   log('debug', 'edited package.json, saving')
 
--- a/generate.js
+++ b/generate.js
@@ -6,7 +6,7 @@ const {
   existsSync
 } = require('node:fs')
 const path = require('node:path')
-const chalk = require('chalk')
+const chalk = require('chalk').default
 const generify = require('generify')
 const parseArgs = require('./lib/parse-args')
 const cliPkg = require('./package')
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -20,28 +20,56 @@ function resolveEnvValue (envPrefix, optionName, type) {
   return envValue
 }
 
+const FALSY_VALUES = new Set(['false', '0'])
+const TRUTHY_VALUES = new Set(['true', '1'])
+
 // Normalize args so they work with util.parseArgs:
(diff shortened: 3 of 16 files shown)


mcollina (approved):
lgtm

bcomnes:
Can we get a release with this pr in it? `fastify-cli@8.0.1` is currently broken due to a a bad pkg-up import.

Tony133:
> Can we get a release with this pr in it? `fastify-cli@8.0.1` is currently broken due to a a bad pkg-up import.

See new release here: https://github.com/fastify/fastify-cli/releases/tag/v8.0.2

PR #917, opened 2026-09-10
fix: make the test runner execute the full suite

mcollina:
Keep the test runner from silently skipping or terminating before tests complete after the glob upgrade.

The runner now awaits glob and test completion, fails on unmatched patterns, serializes shared-fixture tests, and uses in-process execution for the TypeScript ESM suite. This also allows the existing generate-swagger --yaml option through global argument parsing and adds explicit boolean false coverage.

Verified with npm test and npm run lint.


Tony133 (approved):
LGTM

PR #918, opened 2026-09-13
feat: add release workflow

Eomm:
## What

Adds a `release` workflow that publishes `fastify-cli` to npm from the *Actions* tab: pick the bump type (`patch` / `minor` / `major`) and the reusable workflow does the rest.

It delegates to the new reusable workflow proposed in fastify/workflows#237, which bumps the version, commits it as `Bumped vX.Y.Z`, publishes to npm with provenance and creates the GitHub release with generated notes.

image

## Blocked on

- fastify/workflows#237 must be merged and released first. The `uses:` reference is currently pinned to that PR's commit SHA and **must be repinned to the resulting release tag before this PR is merged** (marked with a `TODO` in the file).
- `fastify-cli` must be configured as an npm [trusted publisher](https://docs.npmjs.com/trusted-publishers), since the publish authenticates via OIDC rather than an `NPM_TOKEN` secret.

## Notes

The release workflow does not run the test suite; CI already validates every commit on `main`. There is no `release:build` script in this package, so that step is a no-op here.


Eomm:
The release job now runs inside a `release` [deployment environment](https://github.com/fastify/workflows/pull/237), so only the approved maintainers can actually release.

Before this is merged, the `release` environment has to be created in *Settings -> Environments* with:
- `fastify/release` as a **required reviewer**
- **Prevent self-review** enabled
- deployment branches limited to `main`

Note that an environment that is referenced but never configured is created automatically **with no protection rule at all**, so this step is not optional.

Tony133 (changes requested):
This is blocked on fastify/workflows#237: `uses:` points to a PR commit (`7d11b7f`, already behind the head of fastify/workflows#237).

Next steps: once fastify/workflows#237 is released, repin to the tag (or full SHA) and we're good. On our side we also need the `release` environment in this repo (`fastify/release` as required reviewer, no self-review, deployments only from `main`) and `fastify-cli` set up as npm trusted publisher, otherwise the gate does nothing.

